PR #264, opened 2025-07-28
ISSUE-263 Support admin creds

chibenwa:
Used by Sabre and Twake Mail

Please note that for safety reasons the admin user password needs to be explicitly set in configuration.properties (no defaults) which in turn makes that file compulsory.


calendar-rest-api/src/main/java/com/linagora/calendar/restapi/auth/BasicAuthenticationStrategy.java
```diff
@@ -92,6 +95,10 @@ public Mono<MailboxSession> createMailboxSession(HttpServerRequest httpRequest)
     }
 
     private Mono<Username> authenticate(UserCredentials creds) {
+        if (creds.username.asString().equals(restApiConfiguration.getAdminUsername())
```
vttranlina, line 98: What do you think about using a single username/password in the userDAO/userRepository as admin credentials?
If Sabre only needs one credential to query email users, this is enough.

PR #268, opened 2025-07-30
ISSUE-266 Implement MemoryAlarmEventDAO

hungphan227:


storage-api/src/main/java/com/linagora/calendar/storage/AlarmEvent.java
```diff
@@ -0,0 +1,31 @@
...
+package com.linagora.calendar.storage;
+
+import java.time.Instant;
+
+import org.apache.james.core.MailAddress;
+
+import com.linagora.calendar.storage.eventsearch.EventUid;
+
+public record AlarmEvent(EventUid eventId,
+                         Instant alarmTime,
```
vttranlina, line 28: Per discussion with @hungphan227, 
we will also store:
- Instant eventStartTime 
- boolean recurrenceEvent. 

Then we done all the cases.

storage-api/src/main/java/com/linagora/calendar/storage/AlarmEventDAO.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.linagora.calendar.storage;
+
+import java.time.Instant;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public interface AlarmEventDAO {
+    Mono<AlarmEvent> find(String id);
```
vttranlina, line 27:
```suggestion
    Mono<AlarmEvent> find(String eventUid);
```

for absolute clear

storage-api/src/main/java/com/linagora/calendar/storage/AlarmEventDAO.java
```diff
@@ -0,0 +1,36 @@
...
+
+    Mono<Void> create(AlarmEvent alarmEvent);
+
+    Mono<Void> update(AlarmEvent alarmEvent);
+
+    Mono<Void> delete(String id);
+
+    Flux<AlarmEvent> getByTime(Instant time); // get all alarmEvent with alarmTime <= time
```
vttranlina, line 35:
```suggestion
    Flux<AlarmEvent> listAvailableAlarm(Instant alarmTime, Instant eventStartTime); 
```

vttranlina (approved):

storage-api/src/main/java/com/linagora/calendar/storage/AlarmEvent.java
```diff
@@ -0,0 +1,33 @@
...
+import java.time.Instant;
+
+import org.apache.james.core.MailAddress;
+
+import com.linagora.calendar.storage.eventsearch.EventUid;
+
+public record AlarmEvent(EventUid eventUid,
+                         Instant alarmTime,
+                         Instant eventStartTime,
+                         boolean recurring,
```
vttranlina, line 30: To support the display of email alarm content,
using a `String` `RecurrenceId` is more effective than a boolean.

Please note that information such as the event title, description, and location may be overridden by any recurrence event.

hungphan227: we should keep recurring for case recurring event with only master event

chibenwa:
No objections

Issue #309, opened 2025-08-25
[Resources] PhotoUrlFactory


chibenwa:
Given 

```
  "photos": [
    {
      "url": "https://calendar.linagora.com/linagora.esn.resource/images/icon/laptop.svg",
      "type": "default"
    }
  ]
```

Implement serving that icon: 

 -  put it in the java resources/icons/resources folder of the rest maven plugin the existing openpaas resources cf https://github.com/linagora/linagora.esn.resource/tree/master/frontend/images/icon
 -  we need to implement the route for serving this static content => `GET http://tcalendar.linagora.com:8080/linagora.esn.resource/images/icon/laptop.svg`

Notes: the idea in to eventually serve it via the SPA but for now let's stick to OpenPaaS behaviour.

Note: Be carefull not to do blocking call. Load binary content from classpath upon class instanciation and serve raw binary content from heap memory in runtime. No need caching.

vttranlina:
https://github.com/linagora/twake-calendar-side-service/pull/341

PR #341, opened 2025-09-04
ISSUE-309 [Resources] PhotoUrlFactory

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/309


pom.xml
```diff
@@ -192,6 +192,11 @@
                 <artifactId>icu4j</artifactId>
                 <version>77.1</version>
             </dependency>
+            <dependency>
+                <groupId>io.github.classgraph</groupId>
```
vttranlina, line 196: Guava and Apache Commons Lang3 do not support scanning to list all files in the compiled classpath.


 The [classgraph](https://github.com/classgraph/classgraph) make code much simpler

chibenwa: Thanks for the discovery this little lib look very helpful

PR #352, opened 2025-09-05
Improve doc

chibenwa:
Remains:

 - [ ] Webadmin API documentation
 - [ ] OpenPaaS API documentation
 - [ ] OpenPaaS API auth mecanisms

Side concerns:

 - [ ] Usage of the RabbitMQ management URL shall not be needed
 - [ ] OpenPaaS backnd should not be needed: the proxy mechanism shall be pluggable.


chibenwa:
I am merging this so that e avoid concurrent updates that do not update the doc.

I am still ver happy to do follow up PRs might comment arise.

PR #375, opened 2025-09-10
refactor: replace `SettingBasedResolver.readSavedSettings` with resolveOrDefault to simplify user settings resolution

vttranlina:
Previously, "safe" user settings resolution was duplicated in several places,
making the code hard to maintain. This change introduces `SafeUserSettingsResolver`
as a single entrypoint to centralize resolution and handle default/fallback behavior.


storage-api/src/main/java/com/linagora/calendar/storage/configuration/resolver/SafeUserSettingsResolver.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.james.mailbox.MailboxSession;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linagora.calendar.storage.SimpleSessionProvider;
+import com.linagora.calendar.storage.exception.DomainNotFoundException;
+
+import reactor.core.publisher.Mono;
+
+public class SafeUserSettingsResolver {
```
chibenwa, line 31: I do not get why we would have a safe setting resolver and an unsafe one ?

storage-api/src/main/java/com/linagora/calendar/storage/configuration/resolver/SafeUserSettingsResolver.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public Mono<SettingsBasedResolver.ResolvedSettings> get(MailboxSession session) {
+        return settingsBasedResolver.readSavedSettings(session)
+            .onErrorResume(error -> {
+                logError(session.getUser(), error);
+                return Mono.just(SettingsBasedResolver.ResolvedSettings.DEFAULT);
+            })
+            .switchIfEmpty(Mono.just(SettingsBasedResolver.ResolvedSettings.DEFAULT));
+    }
```
chibenwa, line 55: IMO we would not need this if we update `SettingsBasedResolver::readSavedSettings` instead

storage-api/src/main/java/com/linagora/calendar/storage/configuration/resolver/SafeUserSettingsResolver.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public Mono<SettingsBasedResolver.ResolvedSettings> getWithFallback(Username user, Username secondUser) {
```
chibenwa, line 57: IMO again this can be turned into a top level method within `SettingsBasedResolver`

chibenwa (changes requested):
IMO we do not need the complexity of yet another interface while we could rework `SettingsBasedResolver` to answer all the concerns expressed here.

chibenwa (approved):

PR #395, opened 2025-09-18
ISSUE-394 People search: Fix bug - ensure only user results when objectTypes filter is [`user`]

### LINKED ISSUE #394: people search: contacts can be included when objecttype = user
Given 

```
POST https://twcalendar.linagora.com/api/people/search
{"limit":10,"objectTypes":["user"],"q":"gov"}
```

Then the result includes contacts

```
[
...
    {
        "id": "661a839f-9e90-3aa9-b752-98bfe0127af3",
        "objectType": "contact",
        "photos": [
            {
                "url": "https://twcalendar.linagora.io/api/avatars?email=paul.nationaladmin@govmu.org",
                "type": "default"
            }
        ],
        "emailAddresses": [
            {
                "value": "paul.nationaladmin@govmu.org",
                "type": "Work"
            }
        ],
        "names": [
            {
                "displayName": "paul.nationaladmin@govmu.org",
                "type": "default"
            }
        ],
        "phoneNumbers": []
    }
]
```

Reproduce in unit tests and fix

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/394

### DIFF
--- a/calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/people/search/UserSearchProvider.java
+++ b/calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/people/search/UserSearchProvider.java
@@ -27,7 +27,6 @@ import jakarta.inject.Inject;
 import jakarta.inject.Named;
 
 import org.apache.commons.collections4.CollectionUtils;
-import org.apache.commons.lang3.Strings;
 import org.apache.james.core.Username;
 
 import com.google.common.collect.ImmutableSet;
@@ -38,6 +37,7 @@ import com.linagora.calendar.storage.OpenPaaSUser;
 import com.linagora.calendar.storage.OpenPaaSUserDAO;
 
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 public class UserSearchProvider implements PeopleSearchProvider {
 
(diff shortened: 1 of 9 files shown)


calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/people/search/UserSearchProvider.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public Set<PeopleSearchRoute.ObjectType> supportedTypes() {
+        return OBJECT_TYPES;
+    }
+
+    @Override
+    public Flux<PeopleSearchRoute.ResponseDTO> search(Username username, String query, Set<ObjectType> objectTypesFilter, int limit) {
+        if (CollectionUtils.isEmpty(objectTypesFilter) || objectTypesFilter.contains(ObjectType.CONTACT)) {
```
vttranlina, line 63: ### Why a dedicated `UserSearchProvider`?

Using `contactSearchEngine.autoComplete` for case only `user` leads to the **incorrect limiting bug** (see test case `shouldNotDropValidUsersWhenFilterIsOnlyUser`). Results can be truncated before filtering, causing valid users to be dropped.

### Cons of `UserSearchProvider`

`UserSearchProvider` queries users directly via `userDAO.list()` and applies `matchQuery` in memory. While correct, this may not good for very large user datasets.

### Alternative

1. We could keep using `contactSearchEngine.autoComplete` and set the input limit to `Integer.MAX_VALUE`, then apply `.take(limit)` afterwards in Reactor. However, this would put heavy load on OpenSearch when the dataset is large.
2. Refactor `autoComplete` to support pagination, then we can repeat it

chibenwa (commented):
After much thoughts I agree with the proposed design.

Thanks for paying attention NOT to call OpenSearch and MongoDB in the same call.

vttranlina:
Thanks for push it @chibenwa

PR #434, opened 2025-10-16
[ENHANCEMENT] Back-channel logout: 400 if invalid token

chibenwa:
## Bug Fixes
* Invalid logout token scenarios now properly return HTTP 400 (Bad Request) instead of incorrectly returning HTTP 200 (Success)
* Added validation to ensure required data fields are not empty
* Enhanced error handling with specific error messages for various invalid token conditions, including malformed tokens and missing required claims


chibenwa:
@vttranlina @hungphan227 ?

vttranlina (approved):

PR #543, opened 2025-12-16
fix: Rest API: return JSON Content-Type header for POST /api/configurations

vttranlina:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed HTTP response headers to properly indicate JSON content type in API responses, ensuring client applications correctly identify response format.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


chibenwa (approved):
Is it the only place?

PR #636, opened 2026-03-11
refactor(rest-api): replace legacy OpenPaaS prefixes in primary routes

vttranlina:
Some routes still used legacy OpenPaaS prefixes like:

- `/linagora.esn.dav.import/`
- `/linagora.esn.resource/`

These prefixes came from old OpenPaaS compatibility constraints.

Now that routes are managed centrally in `calendar-side-service`, these prefixes are no longer needed on the main API surface.

## Why

- make the API easier to read
- make frontend client usage clearer

## Notes

Legacy compatibility is still kept only where needed.

## Summary by CodeRabbit

* **New Features**
  * Added a deprecated legacy route bundle to retain backward-compatible endpoints.

* **Bug Fixes**
  * Simplified resource and icon endpoints to friendlier paths (e.g. /api/resources and /images/icon).
  * Updated resource photo URL generation to use the new icon path.
  * Removed the old import-proxy route in favor of the newer import routing.

* **Documentation**
  * API docs updated with new paths, deprecation notes, and import examples.


chibenwa (changes requested):
No breaking change as long as we need to support old calendar and contacts.

vttranlina:
> No breaking change as long as we need to support old calendar and contacts.

Sorry for the unclear description.
This PR is not a breaking change. It still supports the old calendar contact routes.

The backward-compatibility is handled in the [LegacyRoutes](https://github.com/linagora/twake-calendar-side-service/pull/636/changes#diff-dc16ed1424294b88e42a27dcbd5296d10dc9c6b96120f78b6235a7bee1970e3a) class:

hungphan227 (approved):

chibenwa:
> No breaking change as long as we need to support old calendar and contacts.

Make it clearer in the doc then!

Document both routes. Tell which one is deprecated, and link from the old to the new, and from the new to the old.

Eg

> Deprecated in favor of xxxx, kept for OpenPaaS compatibility

or

> Newer implementation of xxx, intended for friendlier paths

Ok?

chibenwa (approved):

PR #643, opened 2026-03-17
ISSUE-611 [Public agenda] Fix AvailableSlotsCalculator: merge by type, intersect across types

vttranlina:
Previous logic unioned all rules (including different types), which incorrectly widened availability.  
`weekly` and `fixed` rules should constrain each other via intersection (union only within the same type).  

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected availability slot calculation to properly intersect overlapping rules instead of merging them, ensuring accurate available time slots when multiple availability rules apply simultaneously.

* **Tests**
  * Added tests validating intersection of weekly and fixed availability rules, including timezone-aware slot computation and mixed rule scenarios.

### DIFF
--- a/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkSlotsRouteTest.java
+++ b/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkSlotsRouteTest.java
@@ -26,7 +26,10 @@ import static io.restassured.http.ContentType.JSON;
 import static net.javacrumbs.jsonunit.assertj.JsonAssertions.assertThatJson;
 
 import java.nio.charset.StandardCharsets;
+import java.time.DayOfWeek;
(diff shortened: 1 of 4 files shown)


chibenwa (changes requested):
Are we sure?

I want to express:
 - I can hold "Calendar front review" every friday afternoon
 - And As I will be in paris Thursday 19/03 I can take those slot on this day too

My feeling is that this is better represented by a union, I fail at seeing where this intersection is coming from

I could very well say "every friday morning and exceptionally all day this friday.

Or also all friday morning and also this friday afternoon.

My feeling is let's not overshoot this, if needed we could add a `overrideRecurringAvailabilitySlotOnSameDay` property to fixed slot availability but as it stands I would prefer keep things simple and keep a union until we get feedback.`

PR #648, opened 2026-03-17
feat - Add more resource icon assets

vttranlina:
ref https://github.com/linagora/twake-calendar-frontend/issues/654


chibenwa (approved):

chibenwa:
> Should we consider changing ResourceIconLoader to load icons from the filesystem (for example, from a path mounted via Docker volume), similar to mail templates?

IMO not required.

PR #660, opened 2026-03-24
chore(amqp): remove unused EventITIPConsumer (for IMIPCALLBACK)

vttranlina:
`IMIPCALLBACK` approach is no longer viable.
EventITIPConsumer-related code has never been used.
Remove it now to reduce code noise and keep the codebase clean before implementing the Async Schedule solution.

## Summary by CodeRabbit

* **Changes**
  * Removed the iTIP event message consumer and its supporting components, including queue registration, reconnection handlers, and the explicit prefetch configuration.

* **Tests**
  * Removed the integration test suite associated with the iTIP event message consumer and updated related expectations.

* **Documentation**
  * Deleted the iTIP prefetch configuration entry from the documentation.


chibenwa (approved):
While we are at it can we:
 - Remove / close IMIP callback work on Sabre side?
 - Close related unmerged Integration Tests ?

Thanks

PR #675, opened 2026-04-08
ISSUE-575 [WS calendarList] Do not throw when calendar base id is a resource

vttranlina:
## Summary by CodeRabbit

* **Improvements**
  * Calendar notification processing now includes a fallback resource resolution mechanism, improving reliability when user lookups are unavailable with enhanced error reporting.


chibenwa (changes requested):
Needs TDD non regression test

chibenwa (approved):

Issue #703, opened 2026-04-22
Multitenant: advertize domain in technical token introspection


chibenwa:
```
GET /api/technicalToken/introspect
```

With the following header:

```
X-TECHNICAL-TOKEN: xyz
```

Will return 200 if the token is valid and 404 otherwise (non-existent or expired)

Body

```
{"domain":"linagora.com", ...}
```

PR #704, opened 2026-04-22
ISSUE-703 Add Domain to technical domain validation response

chibenwa:


chibenwa:
CF https://github.com/linagora/twake-calendar-side-service/issues/703

vttranlina:
What do you think about including `domain` info when signing tokens (`generate`)?

In most flows, this saves one extra query.  
Tradeoff: wider refactor due to signature and call-site updates.

- https://github.com/linagora/twake-calendar-side-service/pull/705

chibenwa:
> Tradeoff: wider refactor due to signature and call-site updates.

That's what I wanted to avoid !

My understanding is that we use technical token mostly in tests and for CardDav domain address member book sync. Is this wrong?

vttranlina:
it is using in 
- sync domain members address book
- Grant/revoke write rights for resource admin
- Update partStat (ACCEPTED/DECLINED) for resource

yes, the benefits are not really that clear

vttranlina (approved):

chibenwa:
Thanks for the detailed clarification

PR #751, opened 2026-05-20
UPGRADE pug4j 2.3.1 -> 2.4.1

chibenwa:
image


chibenwa (commented):
+1

I presume we tested that it had no impact on the display

PR #769, opened 2026-06-01
 ISSUE-759 rest api to get domain settings for authenticated user

hungphan227:


app/src/test/java/com/linagora/calendar/app/restapi/routes/DomainSettingsRouteTest.java
```diff
@@ -0,0 +1,156 @@
...
+            .setBasePath("")
+            .setAuth(auth)
+            .build();
+    }
+
+    @Test
+    void getShouldReturnDefaultsWhenNoSettingsSaved() {
+        given()
+        .when()
+            .get("/api/domainSettings")
```
chibenwa:
```suggestion
            .get("/api/domain/settings")
```

To be more idiomatic ?

storage-api/src/main/java/com/linagora/calendar/storage/DomainSettingsResolver.java
```diff
@@ -0,0 +1,99 @@
...
+    private Boolean configuredResourceSearchEnabled() {
+        return Optional.ofNullable(config.getString("resource.search.enabled"))
+            .map(Boolean::parseBoolean)
+            .orElse(DomainSettings.DEFAULT_RESOURCE_SEARCH_ENABLED);
+    }
+
+    private DefaultCalendarPublicVisibility configuredDefaultCalendarPublicVisibility() {
+        return Optional.ofNullable(config.getString("default.calendar.public.visibility"))
+            .map(DefaultCalendarPublicVisibility::deserialize)
+            .orElse(DomainSettings.DEFAULT_CALENDAR_PUBLIC_VISIBILITY);
```
vttranlina, line 97: The value of `DEFAULT_CALENDAR_PUBLIC_VISIBILITY` is currently `READ`,
while the document says
`default.calendar.public.visibility | Optional. Defaults to private`
Perhaps the value of `DEFAULT_CALENDAR_PUBLIC_VISIBILITY` needs to be updated.

storage-api/src/main/java/com/linagora/calendar/storage/MemoryStorageModule.java
```diff
@@ -81,6 +81,9 @@ protected void configure() {
 
         bind(MemoryTicketStore.class).in(Scopes.SINGLETON);
         bind(TicketStore.class).to(MemoryTicketStore.class);
+
+        bind(MemoryDomainSettingsDAO.class).in(Scopes.SINGLETON);
+        bind(DomainSettingsDAO.class).to(MemoryDomainSettingsDAO.class);
```
vttranlina, line 86: Please install the Guice dependency for the MongoDB module. 
Until the MongoDB implementation is complete, use the in-memory provider to ensure the server starts successfully.

storage-api/src/main/java/com/linagora/calendar/storage/DomainSettingsResolver.java
hungphan227: I think the default value should be read. Doc updated

vttranlina: Before domainSetting story, the default is read, correct?

hungphan227: > Before domainSetting story, the default is read, correct?

It was private

vttranlina: If so, the default should remain `private`, without changing the old logic.

hungphan227: I did not remember what I thought when I set it to private (precisely the value was false at that time). However, for now, I think that in normal circumstances, people should be able to read other's calendars. There is no need to hide.

vttranlina: IMO, when we not sure about something
We should keep old business first
 
@chibenwa any idea?

chibenwa: Yes let's not intruduce beaking changes.

chibenwa (approved):

PR #770, opened 2026-06-02
ISSUE-765 fix(search): stop consuming `request` and `cancel` queues

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/765


hungphan227 (approved):

chibenwa (changes requested):
upgrade-instructions.md explain that one needs to remove those two bindings and delete the related queueu

Failing to do so would result in unbounded rabbitmq messages growth.

chibenwa (approved):

PR #771, opened 2026-06-02
ISSUE-757 perf(people-search): resolve user search mode once

vttranlina:
Resolve user search permissions once per search request before processing contacts. This avoids repeated domain settings lookups for every autocomplete result.


hungphan227 (approved):

chibenwa (approved):
<3

VERY good catch

PR #774, opened 2026-06-03
ISSUE-767 fix(amqp): ignore alarm-only calendar updates

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/767


chibenwa (approved):
Desirable anyway.

Google calendar do not send ITIP upon ALARM update let's do the same!

PR #780, opened 2026-06-04
ISSUE-758 Webadmin routes to get and put domain settings

hungphan227:


app/src/test/java/com/linagora/calendar/app/TwakeCalendarGuiceServerTest.java
```diff
@@ -407,6 +407,25 @@ void shouldExposeCalendarRestApi(TwakeCalendarGuiceServer server) {
...
+            .when()
+            .get("/domains/linagora.com/settings")
+            .then()
+            .extract()
+            .body()
+            .asString();
+
+        assertThatJson(body).isEqualTo("""
+            {
+              "userSearchMode": null,
```
vttranlina, line 422: We have default values, 
and when the admin doesn't configure anything, 
these default values are still being applied in the logic. 
Therefore, the response here must return the default instead of `null`

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/DomainSettingsRoutesTest.java
```diff
@@ -0,0 +1,278 @@
...
+                  "resourceSearchEnabled": false,
+                  "defaultCalendarPublicVisibility": null,
+                  "unknownField": "value"
+                }
+                """)
+            .put("/domains/linagora.com/settings")
+        .then()
+            .statusCode(400);
+    }
+}
```
vttranlina, line 322: Should we implement `PATCH` support for partial resource updates?
For example, a request body containing only `"userSearchMode": "disabled"` should update that specific property while preserving all other existing fields.

app/src/test/java/com/linagora/calendar/app/TwakeCalendarGuiceServerTest.java
hungphan227: IMO admin should work directly with settings in database. If returning default value, the admin may misunderstand that the setting has already been set specifically for the domain

chibenwa: > Therefore, the response here must return the default instead of null

Both opinions are defendable. How about doing something similar to quota where we return both ?

```
GET /domains/linagora.com/settings
{ 
              "userSearchMode": null,
              "resourceSearchEnabled": null,
              "defaultCalendarPublicVisibility": null
              "resolved": { // read only !
                  "userSearchMode": "limited",
                  "resourceSearchEnabled": true,
                  "defaultCalendarPublicVisibility": "private"
             }
}
```

We then get both conciliated: applied value is explicit as well as the configured value.

Opinion?

calendar-webadmin/src/main/java/com/linagora/calendar/webadmin/DomainSettingsRoutes.java
```diff
@@ -0,0 +1,142 @@
...
+    private String putDomainSettings(Request request, Response response) throws JsonExtractException {
+        Domain domain = asDomain(request);
+        DomainSettingsDTO requestDTO = jsonExtractor.parse(request.body());
+        DomainSettings settings = requestDTO.toDomainSettings();
+        domainSettingsDAO.save(domain, settings).block();
+        response.status(HttpStatus.NO_CONTENT_204);
+        return Constants.EMPTY_BODY;
+    }
+
+    private Domain asDomain(Request request) {
```
chibenwa, line 139: Mmm I think we can do better and handle this with a refactoring in webadmin-core. 
I'll propose a refactoring...

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/DomainSettingsRoutesTest.java
chibenwa: Good idea

hungphan227: https://github.com/linagora/twake-calendar-side-service/issues/781

app/src/test/java/com/linagora/calendar/app/TwakeCalendarGuiceServerTest.java
hungphan227: updated

chibenwa (approved):
Fine with me.

vttranlina (approved):

PR #783, opened 2026-06-08
ISSUE-781 Webadmin routes to patch domain settings

hungphan227:


storage-mongodb/src/main/java/com/linagora/calendar/storage/mongodb/MongoDBDomainSettingsDAO.java
```diff
@@ -94,6 +95,54 @@ public Mono<Void> save(Domain domain, DomainSettings settings) {
...
+        if (setDoc.isEmpty() && unsetDoc.isEmpty()) {
+            return Mono.empty();
+        }
+
+        Document update = new Document();
+        if (!setDoc.isEmpty()) {
+            setDoc.append(FIELD_DOMAIN, domain.asString());
+            update.append("$set", setDoc);
+        }
+        if (!unsetDoc.isEmpty()) {
```
chibenwa, line 112: Can we address this finding please

Proposal is to:

Define a 

```
record DomainSettingPatchDocuments(Document setDoc,Document unsetDoc) {
   Document asAppendDocument() { ... }
}
```

Then we could extract each individual field in it's own method, potentially generifying handling of a PatchedValue if relevant.

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/DomainSettingsRoutesTest.java
```diff
@@ -319,4 +319,182 @@ void putShouldReturn400WhenRedundantField() {
...
+                }""");
+    }
+
+    @Test
+    void patchShouldUpdateOnlyTheProvidedField() {
+        domainDAO.add(Domain.of("linagora.com")).block();
+        given().contentType(ContentType.JSON)
+            .body("""
+                {"userSearchMode": "limited", "resourceSearchEnabled": false, "defaultCalendarPublicVisibility": "read"}""")
+            .put("/domains/linagora.com/settings").then().statusCode(204);
```
chibenwa, line 356: That's good but I would like to use "null" to explicitly reset one field

-> Absent -> ignored and unchanged
-> Present and null -> reset


```suggestion
    void patchShouldUpdateOnlyTheProvidedField() {
        domainDAO.add(Domain.of("linagora.com")).block();
        given().contentType(ContentType.JSON)
            .body("""
                {"userSearchMode": null}""")
            .put("/domains/linagora.com/settings").then().statusCode(204);
```

docs/apis/webadmin.md
```diff
@@ -709,6 +709,28 @@ All three fields are required. A `null` value clears the setting for that field,
...
+PATCH /domains/{domain}/settings
+{
+  "userSearchMode": "limited",
+  "resourceSearchEnabled": null
+}
+```
+
+All fields are optional. Only the fields present in the body are updated:
+- A non-null value updates that field to the given value
+- A `null` value clears that field (reverts to system default at resolution time)
```
chibenwa, line 724: Is this tested ?

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/DomainSettingsRoutesTest.java
hungphan227: I have a dedicated test for null case: patchShouldClearFieldWhenNullProvided

docs/apis/webadmin.md
hungphan227: test patchShouldClearFieldWhenNullProvided

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/DomainSettingsRoutesTest.java
chibenwa: Fair I missed it sorry

docs/apis/webadmin.md
chibenwa: Fair I missed it sorry

chibenwa (approved):
Not the way I would have seen the refactoring but I am ok with it

Issue #87, opened 2025-05-05
[STORY] Searching delegated events


chibenwa:
Cuurently broken with OpenPaaS clendar SPA:

![Image](https://github.com/user-attachments/assets/3109d192-06ff-46a5-b1ac-4b0b7aa84415)

New frontend is likely to be needed to fix that bug?

chibenwa:
The events are actually correctly indexed into OpenSearch

But the frontend actually uses the wrong calendar identifiers and finds nothing (FRONTEND FAULT: if I manually fix the identifiers then it works great)

There's a topic for existing events prior delegation. 

GIVEN BOB has events in calendarA
WHEN Alice subscribe to it
THEN Alice should be able to search existing events

(And it is not the case today)

vttranlina:
> Extract the user id from the URL here 5f50a663bdaffe002629099c
> The subscription contains events not yet indexed for that user so index it

How about an alternative approach: instead of adding more consumers and indexing, we search across calendars directly. For example, when Alice performs a search, we first list her subscribed calendars and then run the search in Bob’s account as well.

PR #788, opened 2026-06-10
ISSUE-87 Searching delegated events: resolve calendar search sources from DAV list

vttranlina:
related https://github.com/linagora/twake-calendar-side-service/issues/87

This PR adds a calendar search source resolver for translating client-requested calendar URLs into the actual calendar sources to search. 
It uses the requester calendar list from SabreDAV as the source of truth, so unreadable calendars are filtered and subscribed/delegated mirrors resolve to their source calendars.

## Notes
- The resolver filters unreadable calendars instead of throwing.
- Search is only supported for calendars that are subscribed, delegated, or personal calendars.

### DIFF
--- a/calendar-dav/src/main/java/com/linagora/calendar/dav/CalendarSearchSourceResolver.java
+++ b/calendar-dav/src/main/java/com/linagora/calendar/dav/CalendarSearchSourceResolver.java
@@ -39,6 +39,7 @@ import com.linagora.calendar.storage.CalendarURL;
 import com.linagora.calendar.storage.OpenPaaSId;
 import com.linagora.calendar.storage.OpenPaaSUser;
 
+import it.unimi.dsi.fastutil.Pair;
 import reactor.core.publisher.Mono;
 
 public class CalendarSearchSourceResolver {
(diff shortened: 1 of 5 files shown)


calendar-dav/src/main/java/com/linagora/calendar/dav/CalendarSearchSourceResolver.java
```diff
@@ -0,0 +1,129 @@
...
+            .map(CalendarSearchSourceResolver::toCalendarURL);
+
+    private final CalDavClient calDavClient;
+
+    @Inject
+    public CalendarSearchSourceResolver(CalDavClient calDavClient) {
+        this.calDavClient = calDavClient;
+    }
+
+    public Mono<List<CalendarURL>> resolve(OpenPaaSUser requester, List<CalendarURL> requestedCalendars) {
```
chibenwa:
```suggestion
    public Mono<Map<CalendarURL, CalendarURL>> resolve(OpenPaaSUser requester, List<CalendarURL> requestedCalendars) {
```

So that the caller can take `values()` to get source calendars and inspect the search response to build the href with url the people doing the search supplied ?

chibenwa (approved):

PR #793, opened 2026-06-11
ISSUE-787 add calendarPublicVisibilitySettingEnabled to domain settings

hungphan227:


app/src/test/java/com/linagora/calendar/app/TwakeCalendarGuiceServerTest.java
```diff
@@ -417,18 +417,7 @@ void shouldExposeWebAdminDomainSettings() {
...
-              "resourceSearchEnabled": null,
-              "defaultCalendarPublicVisibility": null,
-              "resolved": {
-                "userSearchMode": "enabled",
-                "resourceSearchEnabled": true,
-                "defaultCalendarPublicVisibility": "private"
-              }
-            }
-            """);
+        assertThat(body).contains("\"userSearchMode\":null");
```
chibenwa, line 421: Why?

docs/apis/webadmin.md
```diff
@@ -696,11 +699,12 @@ PUT /domains/{domain}/settings
...
   "userSearchMode": "limited",
   "resourceSearchEnabled": false,
-  "defaultCalendarPublicVisibility": null
+  "defaultCalendarPublicVisibility": null,
+  "calendarPublicVisibilitySettingEnabled": true
 }
 ```
 
-All three fields are required. A `null` value clears the setting for that field, causing it to fall back to the system default.
+All four fields are required. A `null` value clears the setting for that field, causing it to fall back to the system default.
```
chibenwa:
```suggestion
All fields are required. A `null` value clears the setting for that field, causing it to fall back to the system default.
```

I do not want to see a all five fields are required...

chibenwa (commented):
Stupid

chibenwa (commented):
We need to have changes in the openpaasApi.md doc we need to have this in order to hand over this work to the front.

app/src/test/java/com/linagora/calendar/app/TwakeCalendarGuiceServerTest.java
hungphan227: the purpose of this test is to check if the api is actually exposed, not the content of the response so I made it simple.

chibenwa:
> Stupid

Gosh, very sorry I did not intend indeed to say this here, I believe it's a mistake and that I was typing something else and it ended up in the wrong text field. 

That's not waht I meant, and that's not the way I writte generally on GitHUB.

I'm sorry if this was misinterpreted, it was not intended to describe this work @hungphan227 

Must have been doing to many things at the same time, sorry @hungphan227

app/src/test/java/com/linagora/calendar/app/TwakeCalendarGuiceServerTest.java
chibenwa: Ok yes payload is unmodified but the check made partial so that it do not get modified anytime we add a new domain setting. 

GREAT!

chibenwa (approved):
Thanks!

PR #803, opened 2026-06-16
ISSUE-796 Auto accept resource when organizer is an admin of the resource

hungphan227:


calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
```diff
@@ -177,12 +190,20 @@ private Mono<Void> handleReplyEvent(CalendarResourceMessageDTO message, boolean
             .filter(resource -> !resource.administrators().isEmpty())
             .flatMap(resource -> Mono.just(getOrganizerEmail(message))
                 .filter(eventEmailFilter::shouldProcess)
-                .flatMap(organizerEmail -> sendReplyMail(message, resource.name(), organizerEmail, approved)));
+                .flatMap(organizerEmail -> isOrganizerAdmin(resource, organizerEmail)
```
chibenwa, line 193: I am not really fan of looking up twice if the user is an admin.

How about resolving upfront the OpenPaaSUser associated with organizerEmail and passing it down in place of organizerEmail? Then we could statically check twice if the user is a resource admin without doing some additional DB roundtrips.

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
```diff
@@ -177,12 +190,20 @@ private Mono<Void> handleReplyEvent(CalendarResourceMessageDTO message, boolean
...
+                    .flatMap(ignored -> sendReplyMail(message, resource.name(), organizerEmail, approved))));
     }
 
     private MailAddress getOrganizerEmail(CalendarResourceMessageDTO calendarResourceMessageDTO) {
         VEvent vEvent = GET_FIRST_VEVENT_FUNCTION.apply(calendarResourceMessageDTO.ics());
-        return EventParseUtils.getOrganizer(vEvent).email();
+        return getOrganizer(vEvent).email();
+    }
+
+    private EventFields.Person getOrganizer(VEvent vEvent) {
```
chibenwa: I think we are wrong here...

I do not see what forbids a non admin user to specify an arbitrary ORGANIZER in the event he creates and bypass the admin check.

I would feel more confident if we base the check on the calendar field itself: `calendars/xxx/yyy/zzz.ics` if xxx is the userId on an admin we win. Opinion ?

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
```diff
@@ -139,13 +140,25 @@ public Mono<Void> handleCreateEvent(CalendarResourceMessageDTO message) {
...
+        if (resource.administrators().isEmpty()) {
+            return Mono.just(true);
+        }
+        return isOrganizerAdmin(resource, getOrganizerEmail(message));
+    }
+
+    private Mono<Boolean> isOrganizerAdmin(Resource resource, MailAddress organizerEmail) {
+        Username organizerUsername = Username.fromMailAddress(organizerEmail);
+        return userDAO.retrieve(organizerUsername)
+            .switchIfEmpty(Mono.error(new IllegalStateException("Organizer user not found: " + organizerUsername.asString())))
```
vttranlina: Becareful
method responsibility is `isOrganizerAdmin` 
it should return yes or no
should not throw Exception

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
hungphan227: xxx is resourceId

hungphan227: > xxx is resourceId

in the eventPath of amqp message (exchange: "resource:calendar:event:created")

chibenwa: Ok understood: we do not really have the info of the base calendar here.
And the user is likely free to specify arbitrary organizers which is an issue.
Let's move this in a follow up issue.

chibenwa: CF https://github.com/linagora/esn-sabre/issues/389

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
hungphan227: "looking up twice" --> which operation is twice?

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
hungphan227: It really does not make sense if userDAO could not find organizer. If it actully happen, I think we should throw error to stop the normal flow here.

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
chibenwa: organizer email -> id.

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
chibenwa: Yes but we need to support this failure better.

Please sit with @vttranlina to see this together.

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
hungphan227: not sure what you mean

in the entire flow of handleReplyEvent, calling to db only occur twice:
- resourceDAO.findById
- isOrganizerAdmin(resource, organizerEmail) call userDAO.retrieve(organizerUsername) inside

nothing is duplicated here

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
```diff
@@ -139,13 +140,25 @@ public Mono<Void> handleCreateEvent(CalendarResourceMessageDTO message) {
...
         return resourceDAO.findById(new ResourceId(message.resourceId()))
             .filter(resource -> !resource.deleted())
-            .flatMap(resource -> {
-                if (resource.administrators().isEmpty()) {
-                    return acceptInvite(message, resource);
-                } else {
-                    return sendValidationEmailToAdministrators(message, resource);
-                }
-            });
+            .flatMap(resource -> shouldAutoAccept(message, resource)
```
chibenwa, line 143: shouldAutoAccept call nb 1 wraps a getOrganizerEmail call

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
chibenwa, line 193: isOrganizerAdmin is then called a second time

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
hungphan227: shouldAutoAccept is inside handleCreateEvent method. and handleCreateEvent is to handle message from exchange "resource:calendar:event:created"

calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventResourceHandler.java
hungphan227: no it is called first time. it is in handleReplyEvent method. and handleReplyEvent is to handle message from either exchange "resource:calendar:event:accepted" or "resource:calendar:event:declined"

PR #806, opened 2026-06-16
ISSUE-799 Support indexing resource calendar events for shared calendar search

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/799

The AMQP message payload for resource calendars already works with the `esn-sabre:2.3.1.1` image.

It looks like 
- https://github.com/linagora/esn-sabre/pull/384/
already fixed this, so we do not need any additional code changes.

### DIFF
--- a/calendar-dav/src/test/resources/docker-sabre-dav-setup.yml
+++ b/calendar-dav/src/test/resources/docker-sabre-dav-setup.yml
@@ -43,7 +43,7 @@ services:
 
   sabre_dav:
     hostname: esn_sabre
-    image: linagora/esn-sabre:2.2.1
+    image: linagora/esn-sabre:2.3.1.1
     environment:
       - SABRE_ENV=dev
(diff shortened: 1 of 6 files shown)


calendar-dav/src/test/resources/docker-sabre-dav-setup.yml
```diff
@@ -43,7 +43,7 @@ services:
 
   sabre_dav:
     hostname: esn_sabre
-    image: linagora/esn-sabre:2.2.1
+    image: linagora/esn-sabre:2.3.1.1
```
vttranlina, line 46: I am not using the `esn-sabre:2.3.2` image because that version removes the synchronized scheduling mode.

The side-service test suite still relies on this mode, so using `esn-sabre:2.3.2` would make many unrelated tests fail, outside the scope of this PR.

chibenwa (approved):

PR #811, opened 2026-06-18
fix(opensearch): retry conflicting calendar event updates

vttranlina:
Concurrent event messages can produce optimistic concurrency conflicts in OpenSearch. Retrying allows the sequence-checking script to be reevaluated against the latest document state instead of failing immediately.

### DIFF
--- a/storage-opensearch/src/main/java/com/linagora/calendar/storage/opensearch/OpensearchCalendarSearchService.java
+++ b/storage-opensearch/src/main/java/com/linagora/calendar/storage/opensearch/OpensearchCalendarSearchService.java
@@ -97,6 +97,8 @@ public class OpensearchCalendarSearchService implements CalendarSearchService {
     private static final boolean INDEX_CHECK_SEQUENCE = true;
     private static final CharMatcher QUERY_STRING_CONTROL_CHAR = CharMatcher.anyOf("\"~|*");
     private static final int MAX_SOURCE_CALENDARS_PER_SEARCH = 256;
+    // Retry optimistic concurrency conflicts when concurrent messages update the same event document.
+    private static final int MAX_RETRY_ON_CONFLICT = 3;
 
     private final OpenSearchIndexer indexer;
     private final ReactorOpenSearchClient client;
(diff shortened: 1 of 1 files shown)


storage-opensearch/src/main/java/com/linagora/calendar/storage/opensearch/OpensearchCalendarSearchService.java
```diff
@@ -382,6 +384,7 @@ public Mono<WriteResponseBase> indexWithSequence(DocumentId documentId,
                 .script(UPSERT_WITH_SEQUENCE_SCRIPT.apply(sequence, docMap))
                 .scriptedUpsert(false)
                 .upsert(docMap)
+                .retryOnConflict(MAX_RETRY_ON_CONFLICT)
```
chibenwa, line 387: Hmmm I do hardly get it TBH

Given that SEQUENCE is a monotic integer (can only increase) 
When if we loose once
THEN we are condamned to loose forever

as SEQUENCE never decrease.

Am I missing something?

vttranlina: `retryOnConflict` here is not meant to bypass the SEQUENCE rule.

It handles OpenSearch optimistic concurrency conflicts caused by
concurrent consumer processing.

Example:

Initial state:
- Current document SEQUENCE = 9

Concurrent messages:
- Message A: SEQUENCE = 10
- Message B: SEQUENCE = 11

Race:
- A commits first, document becomes SEQUENCE = 10
- B then fails with an OpenSearch version conflict, even though B is newer

With retryOnConflict:
- B retries against the latest document state: SEQUENCE = 10
- The same script runs again
- Since 11 > 10, B is applied

chibenwa: Understood.

Please confirm that if sequence is lower that stored document the update operation silently return as script condition evaluate to false.

It is thus safe to retry.

Correct ?

vttranlina: Yes, it does re-check the SEQUENCE rule on retry.

chibenwa: > Yes, it does re-check the SEQUENCE rule on retry.

and short-circuit. Correct ?

vttranlina: yes, 
```
if (ctx._source.sequence == null || params.sequence > ctx._source.sequence) {
    ctx._source.putAll(params.doc);
}
```
the condition does not pass and the update becomes a no-op.

chibenwa (approved):

PR #854, opened 2026-06-26
[FIX] Stabilize flaky actionLinksShouldWorkForExternalUser test

chibenwa:
Closes #853

`EventParticipationRouteTest.actionLinksShouldWorkForExternalUser` was flaky on CI, failing with:

```
java.util.NoSuchElementException
	at java.base/java.util.List.getFirst(List.java:823)
	at ...EventParticipationRouteTest.lambda$actionLinksShouldWorkForExternalUser$2(EventParticipationRouteTest.java:306)
```

## Root cause
The test only awaited for the SMTP mail to exist (`hasSize(1)`), then in a soft assertion expected 3 action links and **unconditionally** called `actionLinks.getFirst()`. When the mail body was not yet fully observable, the link list could be empty: the soft assertion merely recorded the `hasSize(3)` failure, but the following `getFirst()` threw a hard `NoSuchElementException` that aborted the test.

## Fix
- Move the recipient check and the action-link extraction inside the `CALMLY_AWAIT` block so they are retried until the mail and its 3 action links are actually available.
- Guard `getHtml` against a not-yet-complete message: return an empty string when the HTML part is not matched yet, so it triggers a retry rather than throwing `IllegalStateException`.

No production code is touched; this is a test-stability change only.

### DIFF
--- a/app/src/test/java/com/linagora/calendar/app/restapi/routes/EventParticipationRouteTest.java
+++ b/app/src/test/java/com/linagora/calendar/app/restapi/routes/EventParticipationRouteTest.java
@@ -290,21 +290,19 @@ class EventParticipationRouteTest {
             .get("/smtpMails")
             .jsonPath();
 
-        CALMLY_AWAIT
-            .atMost(Duration.ofSeconds(10))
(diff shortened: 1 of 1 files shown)


chibenwa (commented):
+1

vttranlina (approved):

PR #855, opened 2026-06-26
chore / test - Fix MongoDB 8.0.6 similar to preprod and prod

vttranlina:


app/pom.xml
```diff
@@ -243,7 +243,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <argLine>-Xms512m -Xmx1024m</argLine>
-                    <forkCount>1</forkCount>
+                    <forkCount>2</forkCount>
```
vttranlina, line 246: I tried setting `forkCount = 1` in the hope that it would make the tests more stable, but it didn't help. 
I reverted the change afterward.

chibenwa (approved):

PR #858, opened 2026-06-26
issues/830 [Team calendar] Add MongoDB entity repository implementation

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/830

### DIFF
--- a/storage-mongodb/src/main/java/com/linagora/calendar/storage/mongodb/MongoDBTeamCalendarRepository.java
+++ b/storage-mongodb/src/main/java/com/linagora/calendar/storage/mongodb/MongoDBTeamCalendarRepository.java
@@ -1,0 +1,177 @@
+/********************************************************************
+ *  As a subpart of Twake Mail, this file is edited by Linagora.    *
+ *                                                                  *
+ *  https://twake-mail.com/                                         *
+ *  https://linagora.com                                            *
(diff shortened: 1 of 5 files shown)


storage-mongodb/src/main/java/com/linagora/calendar/storage/mongodb/MongoDBTeamCalendarRepository.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Override
+    public Flux<TeamCalendar> listByDomain(OpenPaaSId domainId) {
+        return Mono.justOrEmpty(asObjectId(domainId))
+            .flatMapMany(domainObjectId -> Flux.from(collection.find(eq(DOMAIN_ID_FIELD, domainObjectId))))
+            .map(this::fromDocument);
+    }
+
+    @Override
+    public Mono<TeamCalendar> updateDisplayName(TeamCalendarId id, String displayName) {
```
chibenwa, line 128: Hmmm updateDisplayName is harder because we would need to rename the underlying dav collection and its subscriptions I presume.

Maybe we should have it as a split issue.

CF https://github.com/linagora/twake-calendar-side-service/issues/859

chibenwa (approved):

PR #865, opened 2026-06-26
Booking links: reference the booking link in generated event ICS

chibenwa:
Closes #864

First step of #864: events created through a public booking link now carry an `X-OPENPAAS-BOOKING-LINK` ICS property holding the booking link public id.

This ties every booked event back to its originating booking link, which is the prerequisite for the remaining items of the ticket (indexing booking links, exposing an OpenPaaS event search criterion on them, and a user task to clean up all the events of a compromised booking link).

### Changes
- `BookingLinkEventIcsBuilder` adds the `X-OPENPAAS-BOOKING-LINK` property, placed alongside the other `X-OPENPAAS-*` / `X-PUBLICLY-*` properties.
- `BookingLinkReservationService` passes the booking link public id when building the event ICS.
- Unit coverage in `BookingLinkEventIcsBuilderTest` and an assertion in the `BookingLinkReservationRouteTest` integration test.
- CHANGELOG entry.

PR #869, opened 2026-06-27
Booking links: index and expose an event search criterion

chibenwa:
Closes #867

Part 2 of the booking links traceability work (see #864), building on the ICS property added in #865.

## What

Index the originating booking link carried by events created through a public booking link (the `X-OPENPAAS-BOOKING-LINK` property added in #865), and expose a `bookingLink` criterion on the OpenPaaS event search API so all events of a given booking link can be retrieved.

## Changes

- **Indexing**: parse the `X-OPENPAAS-BOOKING-LINK` property on both ingestion paths — the AMQP live-indexing path (`EventFieldConverter` / `EventProperty`) and the CalDAV reindex path (`EventFields.fromVEvent`). The value is carried through `EventFields`, `CalendarEventsDocument`, and mapped as an **indexed** keyword in `CalendarEventIndexMappingFactory` so it is searchable.
- **Search criterion**: add an optional `bookingLink` filter to `EventSearchQuery`, implemented as a term filter in `OpensearchCalendarSearchService` and mirrored in `MemoryCalendarSearchService`.
- **REST exposure (end-user search)**: accept a `bookingLink` field in the `CalendarSearchRoute` search request and expose `x-openpaas-booking-link` in the response data.

## Tests

- `EventFieldConverterTest`: the booking link property is extracted from the jCal message.
- `CalendarSearchServiceContract`: search filters by booking link (runs against both the memory and OpenSearch implementations).
- `CalendarSearchRouteContract`: the REST search filters by `bookingLink` and returns it in the response.

Pre-existing booking events created before this change carry no booking link property and will only be found after a reindex; per #864 this needs no documentation as booking links are not in use yet.

Note: the repository does not currently compile on `calendar-redis` (a `RedisKeyEventDispatcher` constructor mismatch from a snapshot dependency) — this is unrelated to this change and pre-exists on `main`. The touched modules and their dependencies build, and the Docker-free tests pass.

### DIFF
--- a/upgrade-instructions.md
+++ b/upgrade-instructions.md
@@ -53,6 +53,32 @@ index only after confirming no calendar-side-service instance still uses the old
 DELETE /calendar-events
 ```
 
+### Added bookingLinkId field to Calendar Event index mapping
+
+Date: 29/06/2026
+
+We added the `bookingLinkId` field to the OpenSearch calendar event index. This field stores the public
+booking link identifier from the `X-OPENPAAS-BOOKING-LINK` ICS property and powers the `bookingLink`
+criterion of the event search API (`POST /calendar/api/events/search`).
+
+#### Required Actions
+
+**Add the field to your existing index mapping:**
+
+```bash
+PUT /calendar_events/_mapping
+{
+  "properties": {
+    "bookingLinkId": {
+      "type": "keyword",
+      "index": true
+    }
+  }
+}
+```
+
+**Note:** Replace `calendar_events` with your actual index name if different.
+
 ## 2.2.0
 
 ### Removed obsolete search queues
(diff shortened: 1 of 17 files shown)


hungphan227 (approved):

vttranlina (approved):

upgrade-instructions.md
```diff
@@ -53,6 +53,32 @@ index only after confirming no calendar-side-service instance still uses the old
...
+    "bookingLinkId": {
+      "type": "keyword",
+      "index": true
+    }
+  }
+}
+```
+
+**Note:** Replace `calendar_events` with your actual index name if different.
+
```
chibenwa, line 81: Much thanks! I missed this!

PR #873, opened 2026-06-29
ISSUE-841 BookedEventTokenSigner

hungphan227:


calendar-api/src/main/java/com/linagora/calendar/api/BookedEventTokenSigner.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    class Default implements BookedEventTokenSigner {
+
+        interface BookedEventTokenClaim {
+            String PUBLIC_BOOKING_LINK_ID = "publicBookingLinkId";
+            String CALENDAR_ID = "calendarId";
+            String OWNER_ID = "ownerId";
+            String EVENT_ID = "eventId";
```
vttranlina, line 65: I do not understand exactly what value `eventId` contains here.

Is it the eventUID in the ICS, or the eventPathId (resource name, etc.), some value that lets us compute the absolute URL of the event ICS? (This would avoid a report query to find by UID when we update partStat)

hungphan227: It is ics resource id (without .ics). We can say it is eventUid because eventUid's value is the same as ics resource id in this case

vttranlina: How about replace it by `URI calendarEventHref` ?
Example
`abc/xyz/1111.ics` 

I just try to make it easier to use for `CalDavEventRepository` without confuse, no big deal

chibenwa (commented):
No specific objections

PR #874, opened 2026-06-29
ISSUE-842 Update BookedLinkReservationRoute to return jwt

hungphan227:


docs/apis/bookingLink.md
```diff
@@ -418,8 +418,16 @@ Content-Type: application/json
 
 ```
 HTTP/1.1 201 Created
+Content-Type: application/json
+
+{
+  "jwt": "<signed-jwt>"
```
chibenwa: IMO jwt is not explicit enough. I suggest to think about how we could be making the intent more explicit.

`bookingConfirmationToken` would be a good improvment I believe.

vttranlina: updated

vttranlina (approved):

PR #876, opened 2026-06-29
Booking links: admin task to delete all events created by a booking link

chibenwa:
Closes #875

Step 3 of the booking links traceability work (see #864): an admin capability to delete all events that were created through a given booking link. This is convenient for mass clean up, e.g. after a link compromission.

## What

A new webadmin action on the existing booking link admin routes:

```
POST /users/{username}/booking-links/{publicId}?action=deleteEvents[&since=<ISO-8601 instant>]
```

- Schedules an asynchronous task that locates every event of the booking link (via the `X-OPENPAAS-BOOKING-LINK` search criterion indexed in #867) and deletes it from the CalDAV backend.
- Returns `201` with the `taskId`.
- The optional `since` parameter restricts the deletion to events created (or last modified) at or after the given instant, based on `DTSTAMP`. As requested in the issue, the default is no filtering (all events of the link are deleted).

The task exposes `additionalInformation` with `username`, `bookingLinkPublicId`, `since`, `deletedEventCount` and `failedEventCount`.

## Details

- `BookingLinkEventDeletionService` pages through the search results for the booking link's calendar and deletes each matching event, tracking success/failure counts.
- `BookingLinkEventDeletionTask` + `BookingLinkEventDeletionTaskAdditionalInformationDTO` (registered as a webadmin task additional-information module).
- Route wiring in `BookingLinkUserRoutes` (new `TaskManager` + service dependencies).
- Docs (`docs/apis/webadmin.md`) and `CHANGELOG.md` updated.

## Tests

- New `BookingLinkEventDeletionTest` covering deletion, the `since` filter, the empty case, and the `404`/`400` error paths.
- Existing `BookingLinkUserRoutesTest` updated for the new constructor; full class passes (29 tests).

---
*Generated automatically*


calendar-webadmin/src/main/java/com/linagora/calendar/webadmin/service/BookingLinkEventDeletionService.java
```diff
@@ -0,0 +1,170 @@
...
+            .calendars(calendarUrl)
+            .bookingLink(bookingLinkPublicId)
+            .limit(SEARCH_PAGE_SIZE)
+            .offset(offset)
+            .build();
+
+        return searchService.search(query)
+            .collectList()
+            .map(events -> new Page(events, offset + SEARCH_PAGE_SIZE));
+    }
```
chibenwa: We are paging as we are deleting hence the offset with not be stable ! 

Make the page size easy to override and write a test deleting 10 events with a age size of 2: I think there's likely a bug!

Once you confirm this finding please go for the alternative: to base ourself on DAV listing of event, source of truth and likely to be more robust. I really wonder if I do not prefer that option!

vttranlina (approved):

PR #825, opened 2026-06-22
ISSUE-821 JWT link should 404 when organizer deleted meeting

hungphan227:

PR #881, opened 2026-06-30
ISSUE-844 BookedEventCancelRoute

hungphan227:


calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
```diff
@@ -0,0 +1,143 @@
...
+            .flatMap(bookedEventTokenSigner::validateAndExtract)
+            .flatMap(this::deleteBookedEvent)
+            .then(response.status(HttpResponseStatus.NO_CONTENT).send())
+            .onErrorResume(Exception.class, exception -> handleError(response, exception));
+    }
+
+    private Mono<Void> deleteBookedEvent(BookedEvent bookedEvent) {
+        CalendarURL calendarURL = new CalendarURL(new OpenPaaSId(bookedEvent.ownerId()), new OpenPaaSId(bookedEvent.calendarId()));
+        return openPaaSUserDAO.retrieve(new OpenPaaSId(bookedEvent.ownerId()))
+            .flatMap(owner -> calDavClient.deleteCalendarEvent(owner.username(), calendarURL, bookedEvent.eventId()));
```
vttranlina, line 82: Do we need to handle a situation similar to the one we encountered in
- https://github.com/linagora/twake-calendar-side-service/pull/825/

Should we explicitly test what happens when the cancel route is called multiple times?
Should it return an error or support idempotency?

app/src/test/java/com/linagora/calendar/app/restapi/routes/BookedEventCancelRouteTest.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        BookingLink insert(Username username, BookingLinkInsertRequest request) {
+            return bookingLinkDAO.insert(username, request).block();
+        }
+    }
+
+    @RegisterExtension
+    @Order(1)
+    static SabreDavExtension sabreDavExtension = new SabreDavExtension(DockerSabreDavSetup.SINGLETON);
```
vttranlina:
```suggestion
    static SabreDavExtension sabreDavExtension = SabreDavExtension.shared();
```

app/src/test/java/com/linagora/calendar/app/restapi/routes/BookedEventCancelRouteTest.java
```diff
@@ -0,0 +1,242 @@
...
+            .toInstance(new MailTemplateConfiguration("classpath://templates/",
+                MaybeSender.getMailSender("no-reply@openpaas.org"))),
+        binder -> binder.bind(MailSenderConfiguration.class)
+            .toInstance(mailSenderConfigurationFunction.apply(mockSmtpExtension)),
+        binder -> {
+            Multibinder.newSetBinder(binder, GuiceProbe.class)
+                .addBinding()
+                .to(BookingLinkProbe.class);
+
+            binder.bind(MemoryBookingLinkDAO.class).in(Scopes.SINGLETON);
```
vttranlina: Why use memory impl when choice mongodb (that already implemented)?

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
```diff
@@ -0,0 +1,143 @@
...
+import com.linagora.calendar.storage.OpenPaaSUserDAO;
+
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.QueryStringDecoder;
+import reactor.core.publisher.Mono;
+import reactor.netty.http.server.HttpServerRequest;
+import reactor.netty.http.server.HttpServerResponse;
+
+public class BookedEventCancelRoute implements JMAPRoutes {
```
chibenwa: This class seems to duplicate a bunch of code:

`public Stream<JMAPRoute> routes()` IMO we can intruduce an interface called PublicCalendarRoute:

```
public abstract class PublicCalendarRoute extends JMAPRoutes {
    // ... 

    @Override
    public Stream<JMAPRoute> routes() {
        return Stream.of(JMAPRoute.builder()
            .endpoint(endpoint())
            .action((req, res) -> Mono.from(metricFactory.decoratePublisherWithTimerMetric(this.getClass().getSimpleName(), handleRequest(req, res))))
            .corsHeaders());
    }

    protected abstract Mono<Void> handleRequest(req, resp);
}
```

Similarily we can share error responses as static methods of a more global ErrorResponses class:

```
class ErrorResponses {
    public static Mono<Void> doUnauthorizedMissingJWT(HttpServerResponse response) {
        return response.status(HttpResponseStatus.UNAUTHORIZED)
            .headers(JSON_HEADER)
            .sendByteArray(Mono.fromCallable(() -> ErrorResponse.of(401, "Unauthorized", "JWT is missing or invalid").serializeAsBytes()))
            .then();
    }

    public static Mono<Void> doServerError(HttpServerResponse response, Exception exception) {
        return response.status(HttpResponseStatus.INTERNAL_SERVER_ERROR)
            .headers(JSON_HEADER)
            .sendByteArray(Mono.fromCallable(() -> ErrorResponse.of(500, "Server Error", exception.getMessage()).serializeAsBytes()))
            .then();
    }

    // Other meaningfull error responses can be shared here
}
```

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
hungphan227: idempotent test added

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
hungphan227: updated

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
```diff
@@ -0,0 +1,114 @@
...
+            .getOrDefault(BOOKING_CONFIRMATION_TOKEN_PARAM, List.of())
+            .stream()
+            .filter(StringUtils::isNotBlank)
+            .findAny();
+    }
+
+    private Mono<Void> handleError(HttpServerResponse response, Exception exception) {
+        return switch (exception) {
+            case BookedEventTokenClaimException e -> {
+                LOGGER.warn("Invalid booked event token: {}", e.getMessage());
```
chibenwa:
```suggestion
                LOGGER.info("Invalid booked event token: {}", e.getMessage());
```

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
```diff
@@ -0,0 +1,114 @@
...
+                yield ErrorResponseHandler.handle(response, HttpResponseStatus.UNAUTHORIZED, "bookingConfirmationToken is missing or invalid");
+            }
+            case IllegalArgumentException e -> {
+                LOGGER.warn("Bad request for booked event cancellation: {}", e.getMessage());
+                yield ErrorResponseHandler.handle(response, HttpResponseStatus.UNAUTHORIZED, "bookingConfirmationToken is missing or invalid");
```
chibenwa: IllegalArgumentException -> 403 ?

No not really we shall either use a more precsie exception of let it be a 400

> "bookingConfirmationToken is missing

Shouldn't we 404 ? But we are indempotent so that is not true, right?

chibenwa (commented):
Error handling needs refinement,

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
hungphan227: "bookingConfirmationToken is missing" means there is no token in the request

changed 403 -> 400

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
```diff
@@ -98,12 +98,12 @@ private Optional<String> getTokenParameter(HttpServerRequest request) {
     private Mono<Void> handleError(HttpServerResponse response, Exception exception) {
         return switch (exception) {
             case BookedEventTokenClaimException e -> {
-                LOGGER.warn("Invalid booked event token: {}", e.getMessage());
-                yield ErrorResponseHandler.handle(response, HttpResponseStatus.UNAUTHORIZED, "bookingConfirmationToken is missing or invalid");
+                LOGGER.info("Invalid booked event token: {}", e.getMessage());
+                yield ErrorResponseHandler.handle(response, HttpResponseStatus.BAD_REQUEST, "bookingConfirmationToken is missing or invalid");
```
chibenwa: Here I think UNAUTHORIZED was of as the JWT was missing compulsary claims.

chibenwa (approved):

vttranlina (approved):

PR #901, opened 2026-07-03
ISSUE-831 [Team calendar] Add WebAdmin CRUD APIs for entity management

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/831

### DIFF
--- a/calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/TeamCalendarRoutesTest.java
+++ b/calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/TeamCalendarRoutesTest.java
@@ -39,6 +39,7 @@ import org.junit.jupiter.api.extension.RegisterExtension;
 
 import com.linagora.calendar.storage.OpenPaaSDomain;
 import com.linagora.calendar.storage.TeamCalendarInsertRequest;
+import com.linagora.calendar.storage.model.TeamCalendar;
 import com.linagora.calendar.storage.model.TeamCalendarId;
 import com.linagora.calendar.storage.mongodb.DockerMongoDBExtension;
 import com.linagora.calendar.storage.mongodb.MongoDBOpenPaaSDomainDAO;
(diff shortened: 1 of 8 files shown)


calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/TeamCalendarRoutesTest.java
```diff
@@ -0,0 +1,293 @@
...
+            .build();
+    }
+
+    @AfterEach
+    void tearDown() {
+        webAdminServer.destroy();
+    }
+
+    @Test
+    void createShouldPersistTeamCalendar() {
```
chibenwa, line 84: What if I create twice with same data (idempotency ?)
What if I do a second create with different data?)

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/TeamCalendarRoutesTest.java
```diff
@@ -0,0 +1,293 @@
...
+            .body()
+            .asString();
+
+        assertThatJson(response)
+            .inPath("displayName")
+            .isEqualTo("Global Sales");
+    }
+
+    @Test
+    void deleteShouldRemoveTeamCalendar() {
```
chibenwa, line 328: How about delete idempotency ?

calendar-webadmin/src/test/java/com/linagora/calendar/webadmin/TeamCalendarRoutesTest.java
vttranlina: Create is not idempotent. Posting the same payload twice creates two distinct team calendars with different ids
I will add test

chibenwa: Fine, let's just add a test to make this clear!

chibenwa (approved):

PR #912, opened 2026-07-06
ISSUE-911 Sanitize malformed iTIP recipient/sender addresses

chibenwa:
Closes #911

## Problem

`ItipLocalDeliveryConsumer` failed on malformed recipient/sender addresses. When an address was wrapped in angle brackets (e.g. `mailto:<user@gmail.com>` or `mailto:user@gmail.com>`), stripping only the `mailto:` prefix left the brackets in place. Parsing then produced an invalid domain such as `gmail.com>`, throwing:

```
IllegalArgumentException: Domain parts ASCII chars must be a-z A-Z 0-9 - or _ in gmail.com>
```

## Fix

`ItipLocalDeliveryDTO.stripMailto` now trims whitespace and strips surrounding `<`/`>` characters both before and after removing the `mailto:` prefix, handling the various malformed orderings:

- `mailto:user@example.com`
- `mailto:<user@example.com>`
- `mailto:user@example.com>`
- `<mailto:user@example.com>`
- surrounding whitespace

This covers point 1 of the issue (sanitizing `<`/`>`). Point 2 (managing any remaining exception decently) is already handled by the consumer: `consumeMessage` wraps processing in `onErrorResume`, which nacks a genuinely unparseable message to the dead-letter queue instead of crashing the consumer.

## Tests

Added unit tests in `ItipLocalDeliveryDTOTest` covering the sanitization of recipient and sender addresses.

---
*Generated automatically*

### DIFF
--- a/calendar-amqp/src/main/java/com/linagora/calendar/amqp/ItipLocalDeliveryConsumer.java
+++ b/calendar-amqp/src/main/java/com/linagora/calendar/amqp/ItipLocalDeliveryConsumer.java
@@ -227,7 +227,11 @@ public class ItipLocalDeliveryConsumer implements Closeable, Startable {
     // ---- Processing phase --------------------------------------------------------------------
 
     private Mono<Void> processSingleRecipient(ItipLocalDeliveryDTO localDelivery) {
-        Username recipientUsername = Username.of(localDelivery.strippedRecipient());
+        Optional<Username> maybeRecipient = parseRecipient(localDelivery);
+        if (maybeRecipient.isEmpty()) {
+            return Mono.empty();
+        }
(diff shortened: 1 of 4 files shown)


chibenwa (commented):
Nice sanitizing!

Now imagine we get an invalid address, what happens?

Are we handling other recipients? Are we crashing ?

chibenwa (commented):
Ok for this precise instance Invalid addresses are necessarily not local. Can we please manage this terminal error by ignoring the error rather than dead lettering ?

chibenwa (commented):
+1

vttranlina (approved):

PR #915, opened 2026-07-06
ISSUE-913 Booking link: prevent past slots, booking and cancellation

chibenwa:
Closes #913

Booking links currently allow "time travel": past slots are listed, past slots can be booked, and events in the past can be cancelled. This PR makes the backend enforce time consistency.

## Changes

- **Filter out past booking slots** — `BookingLinkSlotsService` now drops any computed slot whose start is before the current time (injected `Clock`). Slot grid alignment on the availability window is preserved (slots are removed, not re-anchored to `now`).
- **Reject reservations in the past** — since a past slot is no longer offered, `BookingLinkReservationService` naturally rejects it with the existing `SlotNotAvailableException` (HTTP 422).
- **Reject cancellations in the past** — `BookedEventCancelRoute` fetches the booked event, and if its start time is already in the past it returns HTTP 422 instead of deleting. A missing event stays a no-op delete, so cancellation remains idempotent.

## Tests

Added `BookingLinkPastManagementRouteTest`, an integration test using a controllable clock (`UpdatableTickingClock`) that covers:
- past slots excluded from the slots listing
- booking a past slot rejected (422), future slot accepted (201)
- cancelling a future event succeeds (204), cancelling an already-started event rejected (422)

Existing `BookedEventCancelRouteTest` still passes.

---
*Generated automatically*

### DIFF
--- a/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkPastManagementRouteTest.java
+++ b/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkPastManagementRouteTest.java
@@ -1,0 +1,291 @@
+/********************************************************************
+ *  As a subpart of Twake Mail, this file is edited by Linagora.    *
+ *                                                                  *
+ *  https://twake-mail.com/                                         *
+ *  https://linagora.com                                            *
+ *                                                                  *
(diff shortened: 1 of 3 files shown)


chibenwa (commented):
+1

PR #917, opened 2026-07-06
ISSUE-834 Support Team Calendar in People Search

hungphan227:


chibenwa (approved):
Fine with me.

FYI I wondex if we shall not create a folllow up issue to:

 - Rely on a event bus for user, team-calendar and resources lifecycle
 - When OpenSearch is configured do indexing through listeners in OpenSearch
 - Alast use OpenSearch based search providers

(The current approach is working for smallish domains but ultimately will likely be limited for biiig domains)

PR #930, opened 2026-07-08
ISSUE-929 Reimport user display name from LDAP upon settings message

chibenwa:
Closes #929

## Context

The issue asks to use the `first_name` / `last_name` payload of the `settings` exchange to update the display name of the registered user, and explicitly accepts *"a systematic reimport of the display name from the LDAP upon `settings` message"* as an alternative.

The shared `TWPCommonSettingsMessage.Payload` (from `tmail-backend`) only exposes `email` and `language`, and its deserializer ignores unknown JSON properties — so `first_name` / `last_name` are not available on the calendar side and cannot be read from this repository. This PR therefore implements the accepted LDAP-reimport approach.

## Changes

- `SaaSUserProvisioner`:
  - resolves first/last name via the `UserNameResolver` (LDAP in production, `Noop` otherwise) when provisioning a **new** user;
  - exposes `reimportDisplayName(OpenPaaSUser)` to refresh the stored display name of an **already registered** user, updating only when it actually changed.
- `CalendarSettingUpdater`: triggers the display-name reimport on **every** `settings` message, independently of whether a language setting is present.

## Tests

Added coverage in `CalendarSettingUpdaterTest` (reimport on rename, reimport even without a language payload, no-op when the resolver returns no name). Full module test suite is green and checkstyle passes.

---
*Generated automatically*

### DIFF
--- a/saas/saas-rabbitmq/src/main/java/com/linagora/calendar/saas/CalendarSettingUpdater.java
+++ b/saas/saas-rabbitmq/src/main/java/com/linagora/calendar/saas/CalendarSettingUpdater.java
@@ -77,12 +77,16 @@ public class CalendarSettingUpdater implements TWPSettingsUpdater {
 
     @Override
     public Mono<Void> updateSettings(TWPCommonSettingsMessage message) {
-        return message.languageSettings()
-            .map(incomingLanguageSetting -> resolveUsername(message)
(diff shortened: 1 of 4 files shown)


chibenwa (commented):
+1

PR #933, opened 2026-07-08
ISSUE-931 Add optional color property to booking links

chibenwa:
Closes #931

## What

Adds an optional `color` property to booking links.

- Stored as an `Optional<String>` hex color (`#RRGGBB`) in `BookingLink`, persisted by both the Mongo and in-memory DAOs. Retro-compatible: links created before this change simply have no stored color.
- **Always exposed** by every booking-link representation, defaulting to `#6B4ECC` when unset:
  - private REST API `BookingLinkDTO` (`GET /api/booking-links` list + single)
  - public REST API `BookingLinkSlotsResponse` (`GET /api/booking-links/{id}/slots`)
  - webadmin `/users/{username}/booking-links` routes
- Accepted (optionally) on create (`POST`) and patch (`PATCH`), for both the end-user and webadmin APIs. Blank values are ignored; on patch, `null`/blank removes the stored color (default then applies).
- Input is validated as a `#RRGGBB` hex string (`BookingLinkColor`), returning `400` otherwise.

## Docs

Updated `docs/apis/bookingLink.md` and `docs/apis/webadmin.md` (data model, create/patch request bodies, and sample responses).

## Tests

- `BookingLinkDAOContract` gains color coverage (persist / default-to-empty / update / remove), run against the memory and Mongo DAOs.
- Route tests updated so every booking-link JSON body includes the always-present `color`, plus new create/get/patch tests (custom color, default, invalid-hex `400`, removal).
- Verified locally: full compilation of all modules, `MemoryBookingLinkDAOTest` (37) and `MongoDBBookingLinkDAOTest` (37) green, and `BookingLinkGetRouteTest` (11) green. The remaining container-based route tests were not run locally due to a full disk on the environment; they will run in CI.

---
*Generated automatically*


chibenwa (commented):
+1

storage-api/src/main/java/com/linagora/calendar/storage/booking/BookingLinkColor.java
```diff
@@ -0,0 +1,44 @@
...
+ ********************************************************************/
+
+package com.linagora.calendar.storage.booking;
+
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+import com.google.common.base.Preconditions;
+
+public class BookingLinkColor {
```
hungphan227: Should it be BookingLinkColorUtil? "BookingLinkColor" is name for model

chibenwa:
```suggestion
public class BookingLinkColorUtil {
```

chibenwa (commented):
+1

hungphan227 (approved):

PR #936, opened 2026-07-08
ISSUE-934 Tolerate version conflicts on calendar event delete-by-query

chibenwa:
Closes #934

## Problem

`EventIndexerConsumer` logs `Failed to delete eventUid ... from calendarURL ...` with an OpenSearch `409 Conflict` / `version_conflict_engine_exception`.

An OpenSearch `_delete_by_query` runs in two phases: it first snapshots the documents matching the query, then bulk-deletes them using the sequence numbers captured during the snapshot. When a concurrent (or reordered) message re-indexes one of those documents in between, its sequence number changes, so the bulk delete is rejected with a version conflict (HTTP 409) and the whole delete aborts. The event is therefore left in the search index and the failure is only logged as a `WARN`.

## Fix

`OpensearchCalendarSearchService` now issues the delete-by-query with `conflicts=proceed` instead of aborting, and retries a bounded number of times while version conflicts remain, so the concurrently-rewritten documents are still removed. Residual conflicts (a document under sustained writes) are left for a later message to clean up rather than looping indefinitely.

This applies to all three delete paths: `delete`, `deleteAll` and stale-occurrence pruning.

## Test

Added `deleteShouldNotFailWhenEventIsConcurrentlyReindexed`, which interleaves continuous re-indexing with deletes of the same event. It reliably reproduces the exact `version_conflict_engine_exception` from the issue (verified: the test fails with `conflicts=abort`) and passes with the fix.

---
*Generated automatically*

### DIFF
--- a/storage-opensearch/src/test/java/com/linagora/calendar/storage/opensearch/OpensearchCalendarSearchServiceTest.java
+++ b/storage-opensearch/src/test/java/com/linagora/calendar/storage/opensearch/OpensearchCalendarSearchServiceTest.java
@@ -18,6 +18,7 @@
 
 package com.linagora.calendar.storage.opensearch;
 
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
 
 import java.util.List;
@@ -43,6 +44,8 @@ import com.linagora.calendar.storage.eventsearch.CalendarEvents;
 import com.linagora.calendar.storage.eventsearch.CalendarSearchService;
 import com.linagora.calendar.storage.eventsearch.CalendarSearchServiceContract;
 import com.linagora.calendar.storage.eventsearch.EventSearchQuery;
+
+import reactor.core.publisher.Flux;
 
 public class OpensearchCalendarSearchServiceTest implements CalendarSearchServiceContract {
     @RegisterExtension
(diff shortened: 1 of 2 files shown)


chibenwa (commented):
+1

storage-opensearch/src/test/java/com/linagora/calendar/storage/opensearch/OpensearchCalendarSearchServiceTest.java
```diff
@@ -214,4 +217,26 @@ void searchShouldReturnEmptyResultWhenQueryStringQueryDoesNotMatch(String search
...
+
+        CalendarEvents events = CalendarEvents.of(event);
+
+        Flux<Void> reindexing = Flux.range(0, 300)
+            .flatMap(i -> testee().reindex(events), 8);
+
+        Flux<Void> deleting = Flux.range(0, 100)
+            .concatMap(i -> testee().delete(event.calendarURL(), event.uid()));
+
+        assertThatCode(() -> Flux.merge(reindexing, deleting).blockLast())
```
vttranlina, line 239: I think this assertion is not enough.
It currently only asserts that no error is thrown.
I think coverage would be better if we assert that the delete function succeeds.

vttranlina (approved):

PR #940, opened 2026-07-08
ISSUE-939 Booking link: email upon bookee cancellation

chibenwa:
Closes #939

When a booker cancels their reservation on a booking link, we now notify both parties:

- The booker receives a friendly "booking request cancelled" acknowledgement email (new event-booking-request-cancelled template, translated in en/fr/ru/vi).
- The booking link owner (organizer) receives an ICS cancellation email (METHOD:CANCEL, STATUS:CANCELLED, bumped SEQUENCE) reusing the existing event-cancel template.

This mirrors the booking creation flow (acknowledgement to booker + ICS proposal to organizer). Notifications are best-effort: failures are logged and swallowed so they never break the cancellation itself. All data needed for the emails is extracted from the ICS fetched before deletion, since the cancellation token only carries identifiers.

Note: deleting the event on the organizer's calendar also emits a standard iTIP CANCEL to the attendee (pre-existing platform behaviour common to all event deletions), so the booker may additionally receive that raw ICS cancel on top of the friendly acknowledgement. The two emails required by the issue are the ones added here.

Tested end-to-end in BookedEventCancelRouteTest#cancellationShouldNotifyBookerAndOrganizer.

---
*Generated automatically*


chibenwa (commented):
+1

I'd like to have a page in /docs mail.md that this in a brief table all cases the calendar sends mail and to who.

chibenwa (commented):
+1

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookedEventCancelRoute.java
```diff
@@ -90,20 +99,63 @@ protected Mono<Void> handleRequest(HttpServerRequest request, HttpServerResponse
...
+        Mono<Void> deleteEvent = calDavClient.deleteCalendarEvent(owner.username(), calendarURL, eventId);
+
+        return fetchCalendarEvent(owner.username(), calendarURL, eventId)
+            .flatMap(davCalendarObject -> rejectIfAlreadyStarted(davCalendarObject.calendarData(), eventId)
+                .then(deleteEvent)
+                .then(notifyCancellation(owner, davCalendarObject.calendarData()))
+                .thenReturn(Boolean.TRUE))
             // A missing event yields an empty Mono and stays a no-op delete, keeping cancellation idempotent.
-            .map(davCalendarObject -> EventParseUtils.getStartTime(EventParseUtils.getFirstEvent(davCalendarObject.calendarData())).toInstant())
+            .switchIfEmpty(deleteEvent.thenReturn(Boolean.TRUE))
```
hungphan227: If fetchCalendarEvent returns empty, it means that the event does not exist. So why do we have to delete an event that does not exist?

chibenwa (commented):
@vttranlina is right: the bookee receives duplicated email:
 - Regular ITIP cancel
 - Booking request cancelled

We can very likely keep the current bookee perspective which means we just need the mail to the owner in case of cancelation.

chibenwa (commented):
+1

hungphan227 (approved):

PR #944, opened 2026-07-09
ISSUE-923 Booking link event invitation do not have blue bar

chibenwa:
Closes #923

## Diagnosis

When Bob books Alice through a booking link, Alice receives the event proposal email built by `PublicAgendaProposalNotifier`. That email attaches the ICS produced by `BookingLinkEventIcsBuilder`, i.e. the very calendar object that gets `PUT` on the CalDAV server — and a stored calendar object must not carry a `METHOD` property.

The MIME part was declared as `text/calendar; charset=UTF-8; method=REQUEST`, but the ICS body itself had no `METHOD:REQUEST` line, which is what mail clients read to decide whether the mail is an iTIP invitation. Hence no blue bar.

The very same problem affects the `event-booking-confirmed` email emitted by `EventMailHandler` when the booking is auto-accepted: it forwards the calendar received over AMQP as-is.

## Fix

Add `CalendarUtil.withMethod(calendar, method)`, returning a copy of the calendar carrying the supplied iTIP `METHOD`, and use it whenever a calendar is turned into a `text/calendar` MIME part:

- `PublicAgendaProposalNotifier` (proposal email to the booking link owner) via the new `BuildResult.icsBytes(Method)`,
- `EventMailHandler.EventMessageGenerator.createAttachments` (invite, update, cancel, reply, counter and booking confirmed emails). For the regular iTIP flows the calendar already carries the METHOD Sabre put in it, so this is a no-op there.

`BookedEventCancelled` now reuses that helper rather than open-coding it.

The stored calendar object is left untouched: `BuildResult.icsBytes()` (used for the CalDAV import) still produces a METHOD-less ICS.

## Tests

- `CalendarUtilTest`: the helper adds, overrides, and does not mutate its input.
- `BookingLinkEventIcsBuilderTest`: the ICS pushed to DAV has no METHOD, the one asked with a METHOD has it.
- `BookingLinkReservationRouteTest#proposalEmailShouldCarryARequestIcs`: end-to-end, the ICS attached to the organizer email contains `METHOD:REQUEST`. Verified to fail before the fix.
- `EventPublicAgendaEmailConsumerTest`: decoded `text/calendar` part of the booking confirmed email contains `METHOD:REQUEST`.

---
*Generated automatically*

### DIFF
--- a/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkReservationRouteTest.java
+++ b/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkReservationRouteTest.java
@@ -366,6 +366,34 @@ class BookingLinkReservationRouteTest {
     }
 
     @Test
+    void proposalEmailShouldCarryARequestIcs(TwakeCalendarGuiceServer server) {
+        BookingLink inserted = insertActiveBookingLink(server);
+        String slotStartUtc = getAvailableSlots(inserted.publicId()).getFirst();
+
+        given()
+            .pathParam("bookingLinkPublicId", inserted.publicId().value())
+            .body("""
+                {
+                  "startUtc": "%s",
+                  "creator": {
+                    "name": "BOB",
+                    "email": "creator@example.com"
+                  },
+                  "eventTitle": "30-min intro call"
+                }
+                """.formatted(slotStartUtc))
+        .when()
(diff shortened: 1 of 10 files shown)


chibenwa (commented):
+1

hungphan227 (approved):

chibenwa (commented):
Please rebase this work on linagora/main

vttranlina (approved):

PR #946, opened 2026-07-09
ISSUE-945 WS notification upon booking link changes

chibenwa:
Closes #945

Implements the payload-free, convergent notification agreed on in the issue: any booking link CRUD emits a `BookingLinkStateChangedEvent` on the owner's `UsernameRegistrationKey`, delivered over the WebSocket as:

```json
{"bookingLinkStateChanged": true}
```

The front re-fetches `GET /api/booking-links` on reception. No delta is carried, so a dropped message is recovered by the next mutation and no ordering has to be maintained. No client-side `register` is needed: the notification rides the `calendarList` default subscription, which is registered unconditionally at connection time on the very same key.

## Changes

- `BookingLinkStateChangedEvent` (storage-api): payload-free, `isNoop()` returns `false`.
- `EventBusBookingLinkDAO`: a `BookingLinkDAO` decorator dispatching on `insert` / `update` / `delete` / `resetPublicId`. Being at the DAO level it covers both the REST routes and `BookingLinkUserRoutes` (webadmin), and since every mutating method takes the owner `Username`, an administrator writing on behalf of a user notifies that user rather than himself. Bound through `@Provides` in `MemoryStorageModule` and `MongoDBStorageModule`, both of which are only installed where `EventBus` is bound; the DAO contract tests keep instantiating the raw DAOs.
- `CalendarEventSerializer`: added `BookingLinkStateChangedDTO` to `@JsonSubTypes`, `toDTO` and `fromDTO`. Without this the event would work on `InVMEventBus` and be silently dropped once Redis is enabled.
- `DefaultWebSocketNotificationListener`: handles the new event and emits `BookingLinkStateChangedMessage`.
- `docs/apis/websocket.md`: documents the push, and states that clients should debounce bursts.

Semantics: spurious events are acceptable, missing ones are not. Deleting an already deleted link emits an event with nothing having changed, which is harmless given the re-fetch semantic. `POST /api/booking-links/{id}/book` does not mutate the link record and emits nothing — the resulting calendar change already reaches the owner through the existing calendar event push.

## Tests

- `CalendarEventSerializerTest`: serialization round-trip for the new DTO.
- `WebsocketRouteTest` (Redis + MongoDB): the frame arrives on the default subscription without any `register`, upon creation, upon deletion, and upon a REST `PATCH`; a user does not receive the notifications of another user.

All four WebSocket tests and the serializer tests pass locally; checkstyle is clean.

---
*Generated automatically*

### DIFF
--- a/app/src/test/java/com/linagora/calendar/app/restapi/routes/WebsocketRouteTest.java
+++ b/app/src/test/java/com/linagora/calendar/app/restapi/routes/WebsocketRouteTest.java
@@ -29,6 +29,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
@@ -61,6 +62,7 @@ import com.google.inject.Provides;
 import com.google.inject.Singleton;
 import com.google.inject.multibindings.Multibinder;
 import com.linagora.calendar.app.AppTestHelper;
+import com.linagora.calendar.app.BookingLinkProbe;
 import com.linagora.calendar.app.TwakeCalendarConfiguration;
 import com.linagora.calendar.app.TwakeCalendarExtension;
 import com.linagora.calendar.app.TwakeCalendarGuiceServer;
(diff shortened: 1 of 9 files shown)


chibenwa (commented):
+1

vttranlina (approved):

chibenwa (commented):
+1 thanks for the fixup @vttranlina

PR #903, opened 2026-07-03
feat: add audit log consumer for Sabre CRUD operations

Arsnael:
Closes #597

Listen to all Sabre RabbitMQ exchanges (calendar, event, contact, addressbook, subscription CRUD) and log audit events via a dedicated 'audit' logger.

---
*Generated automatically*

PR #951, opened 2026-07-10
ISSUE-950 AuditTrail: log connectedUser

chibenwa:
Closes #950

## Why

[esn-sabre](https://github.com/linagora/esn-sabre/pull/419/commits/21810f6b8082e9af4406a0a62ef9924e96eacec0) now injects a `connectedUser` field into the AMQP payloads it publishes: the principal URI of the user who *actually performed* the action. That principal may differ from the resource owner (delegation, admin impersonation, technical token), which is precisely what makes it worth auditing -- the audit log introduced by linagora/twake-calendar-side-service#903 currently only reports the owner deduced from the resource path.

## What

- `EventAuditLogConsumer` now extracts `connectedUser` from the payload and adds it as a `connectedUser` parameter of the audit log entry, alongside the existing `path`.
- The field is optional: when it is absent (older Sabre), `null`, or blank, no parameter is emitted and nothing fails. `username` (the owner) and the human readable message are unchanged.

## Tests

Unit tests in `EventAuditLogConsumerTest` cover the field being present, absent, `null`, blank, and a non-JSON payload, plus the delegation case where `connectedUser` (the delegate) differs from the owner derived from `eventPath`.

Integration coverage of the field on the wire lives in linagora/twake-calendar-integration-tests#293.

---
*Generated automatically*

### DIFF
--- a/calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventAuditLogConsumer.java
+++ b/calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventAuditLogConsumer.java
@@ -252,9 +252,13 @@ public class EventAuditLogConsumer implements Closeable, Startable {
     public static Optional<String> extractConnectedUser(String body) {
         try {
             JsonNode root = MAPPER.readTree(body);
-            return Optional.ofNullable(root.path("connectedUser").asText(null))
+            if (!root.has("connectedUser")) {
+                return Optional.empty();
+            }
+            return Optional.ofNullable(root.get("connectedUser").asText(null))
                 .filter(user -> !user.isBlank());
         } catch (Exception e) {
+            LOGGER.error("Error when extracting connectedUser from audit log event", e);
             return Optional.empty();
         }
     }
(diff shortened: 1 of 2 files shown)


calendar-amqp/src/main/java/com/linagora/calendar/amqp/EventAuditLogConsumer.java
```diff
@@ -243,6 +244,21 @@ public static Optional<String> extractOwner(String body) {
...
+     * It may differ from the resource owner (delegation, admin impersonation), which is what
+     * makes it worth auditing. Older Sabre versions do not emit it: absence is not an error.
+     */
+    public static Optional<String> extractConnectedUser(String body) {
+        try {
+            JsonNode root = MAPPER.readTree(body);
+            return Optional.ofNullable(root.path("connectedUser").asText(null))
+                .filter(user -> !user.isBlank());
+        } catch (Exception e) {
+            return Optional.empty();
```
chibenwa, line 262: Handle  missing connectedUser with an explicit check. 
Do not swallow exceptions. If something bad happen ERROR log but handle missing with an explicit upfront check

chibenwa, line 262: Handle missing connectedUser with an explicit check.
Do not swallow exceptions. If something bad happen ERROR log but handle missing with an explicit upfront check

PR #959, opened 2026-07-16
ISSUE-958 Booking link: extraAttendee field

chibenwa:
Closes #958

Booking links now carry an optional `extraAttendees` field: a list of OpenPaaS ids of registered users, so a single link can be handed over for a meeting that needs several people (the sales guy, the presales engineer and the project manager of the ticket's example).

The field is surfaced on both the OpenPaaS (`POST`/`GET`/`PATCH /api/booking-links`) and webadmin (`/users/{username}/booking-links`) APIs. It is empty when omitted on creation and when absent from the database, and is omitted from responses when empty — consistent with how `availabilityRules`, `name` and `description` already behave.

It changes the booking link in two ways:

- **Availability is the intersection** of the availability rules, the owner availability, and each extra attendee availability.
- **Booked events invite the extra attendees**, with `PARTSTAT=NEEDS-ACTION`: unlike the booker, they were never asked, so they answer through the regular iTIP flow.

Working hours in the extra attendees own timezone are left out, as the ticket suggests.

## Points worth your attention

**Attendee availability is whatever the DAV server shows the owner.** The free-busy `REPORT` is issued impersonating the booking link owner, so Sabre decides what the owner may see. I initially wrote a test asserting that a *non*-delegated calendar contributes no busy intervals, and it failed: Sabre answers free-busy on another user's calendar without any explicit delegation. So "public calendar and read delegation" is enforced by the DAV server rather than by this code — which I think is the right split, but it means the effective visibility rule is Sabre's, not ours.

**An unreadable attendee calendar is treated as free** (warning logged) rather than failing the whole `/slots` and `/book` call. Offering a slot an extra attendee will decline seemed clearly less bad than a booking link that 500s permanently the day someone revokes a share. Same reasoning for an extra attendee whose account is deleted after link creation: they are skipped at booking time instead of making the link unbookable forever. Both are tested. Happy to make either strict if you would rather fail loudly.

**Validation** rejects unknown ids and the owner themselves (400), caps the list at 20, and dedupes. An extra attendee who also happens to be the booker keeps a single `ATTENDEE` line rather than getting a duplicate with a conflicting `PARTSTAT`.

## Tests

All green, checkstyle clean.

- `BookingLinkDAOContract`: 5 new cases (persist, default-empty, patch, remove, keep), so both Memory and MongoDB — 42 each.
- `BookingLinkExtraAttendeeResolverTest` (new), `BookingLinkEventIcsBuilderTest`: validation, lenient resolution, ICS attendees and dedup.
- App routes: create/get/patch/slots/reservation — 185 booking-link tests pass, including slot intersection, the unreadable-calendar fallback, and the deleted-attendee booking.
- Webadmin: `BookingLinkUserRoutesTest` — 34 pass.

Docs (`docs/apis/bookingLink.md`, `docs/apis/webadmin.md`) and `CHANGELOG.md` updated.

---
*Generated automatically*

### DIFF
--- a/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkPatchRouteTest.java
+++ b/app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkPatchRouteTest.java
@@ -31,6 +31,7 @@ import java.time.DayOfWeek;
 import java.time.Duration;
 import java.time.LocalTime;
 import java.time.ZoneId;
+import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 
@@ -127,6 +128,12 @@ class BookingLinkPatchRouteTest {
             .build();
 
         calDavClient = new CalDavClient(sabreDavExtension.dockerSabreDavSetup().davConfiguration(), TECHNICAL_TOKEN_SERVICE_TESTING);
+    }
+
+    private OpenPaaSUser newProvisionedUser(TwakeCalendarGuiceServer server) {
+        OpenPaaSUser user = sabreDavExtension.newTestUser();
+        server.getProbe(CalendarDataProbe.class).addUserToRepository(user.username(), PASSWORD);
+        return user;
     }
 
     @Test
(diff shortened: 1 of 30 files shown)


app/src/test/java/com/linagora/calendar/app/restapi/routes/BookingLinkPatchRouteTest.java
```diff
@@ -388,6 +395,67 @@ void shouldRemoveAvailabilityRulesWhenSetToNull() {
...
+                """)
+        .when()
+            .patch("/api/booking-links/" + inserted.publicId().value())
+        .then()
+            .statusCode(HttpStatus.SC_NO_CONTENT);
+
+        BookingLink updated = bookingLinkProbe.findBookingLink(openPaaSUser.username(), inserted.publicId());
+        assertThat(updated.extraAttendees()).isEmpty();
+    }
+
```
chibenwa, line 438: Can we add a test when we try to add the owner as extraAttendee on PATCH ? 

I could not see the same test than for createRoute

Also lets make sure invalid/not found ids result in 400 for PATCH

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookingLinkExtraAttendeeResolver.java
```diff
@@ -0,0 +1,85 @@
...
+                    LOGGER.warn("Skipping extra attendee {}: no such user anymore", id.value()))));
+    }
+
+    public Mono<Void> validate(Username owner, List<OpenPaaSId> extraAttendees) {
+        if (extraAttendees.isEmpty()) {
+            return Mono.empty();
+        }
+        Mono<OpenPaaSId> ownerId = openPaaSUserDAO.retrieve(owner)
+            .map(OpenPaaSUser::id)
+            .cache();
```
chibenwa: Why do we need to cache ?

```
openPaaSUserDAO.retrieve(owner)
            .map(OpenPaaSUser::id)
            .flatMap(ownerId -> ...)
```

chibenwa (commented):
I'm OK with the code but I wish the feature to be more future proof.

Calendaring might have complex logic like optional participants, substitutes (bob OR michael, but 1 of them) 

Thus we can not keep a flat field for extra participants.

My proposal is to expend it into

```
{
  "extraAttendees": {
    "and": [
      { "participant": "participant1" },
      { "participant": "participant2" }
    ]
  }
}
```

Which would enable in the future to write:


```
{
  "extraAttendees": {
    "and": [
      { "participant": "participant1" },
      { "or: [
        { "participant": "participant2" },
        { "participant": "participant3" }
       }
    ]
  }
}
```

With exactly that semantic 1 and + participant leaf which express exactly today behaviour, that will remain unchanged in the resolvers.

DB wize I'd like the extraAttendees DB field to be structured so that we can model architrary complex attendee trees, but surfacing today flat list.

We are planning for future evolution here, not implementing ANY of that complexity now (appart a simple ser/deser logic on webadmin+ openpaas API and the mongo collection)

chibenwa (commented):
+1

PR #966, opened 2026-07-17
ISSUE-965 URL encode the location in mail map links

chibenwa:
Closes #965

## Problem

The mail templates built the "See in Map" link by interpolating the raw location straight into the query string:

```pug
a(href=`https://www.openstreetmap.org/search?query=${content.event.location.value}`)
```

Pug HTML-escapes attributes but does not URL-encode them, so a location like `37 Rue Pierre Poli, 92130 Issy-les-Moulineaux` ended up in the `href` with literal spaces. Mail clients strip those spaces, the query becomes `37RuePierrePoli,92130Issy-les-Moulineaux`, and OpenStreetMap finds nothing.

## Fix

`LocationModel` now exposes a percent-encoded `urlEncodedValue` next to the raw `value`, and the ten templates that render a map link use it to build the URL. The location text shown to the reader still uses the raw `value`.

Encoding is `URLEncoder.encode(value, UTF_8)` with `+` rewritten to `%20`, so spaces become `%20` rather than `+` and non-ASCII locations are encoded as UTF-8 (`Café` → `Caf%C3%A9`).

The issue's address now yields:
`https://www.openstreetmap.org/search?query=37%20Rue%20Pierre%20Poli%2C%2092130%20Issy-les-Moulineaux`

## Tests

New `LocationModelTest` covers the encoding, including the exact address from the issue. The existing render tests (`HtmlBodyRendererTest`, `HtmlEmailInviteRenderTest`, and the other `Html*RenderTest` classes) had their expected HTML updated to the encoded URL — `HtmlBodyRendererTest` goes through the real `LocationModel` and renderer, so it confirms the fix end to end. All pass.

---
*Generated automatically*


chibenwa (commented):
+1 !

PR #972, opened 2026-07-20
ISSUE-833 [Team calendar] Add integration tests for Team Calendar flows

vttranlina:


app/src/test/java/com/linagora/calendar/app/restapi/routes/TeamCalendarFlowIntegrationTest.java
```diff
@@ -0,0 +1,413 @@
...
+        findVisibleTeamCalendar(bob, teamCalendarId);
+
+        // Then the team calendar is not delegated to the non-member
+        assertThat(calDavClient.findUserCalendarList(nonMember).block().calendars().entrySet())
+            .noneMatch(entry -> isTeamCalendar(entry, teamCalendarId));
+    }
+
+    @Test
+    void deletingTeamCalendarShouldCleanupMetadataAndDavData(TwakeCalendarGuiceServer server) {
+        // Given Bob is a delegated member of a team calendar
```
vttranlina, line 171: There is a related issue when deleting a team-calendar through WebAdmin.

After that, members get errors when reporting the calendar list.

The cause is that deleting a team-calendar through WebAdmin only removes metadata from the team-calendars table, while the calendars in calendarinstances are not cleaned up correctly.

I am implementing an approach where deletion through WebAdmin calls sabre to clean them up.

chibenwa (approved):

PR #974, opened 2026-07-20
ISSUE-973 Booking acknowledgement email: link to the reservation page

chibenwa:
Closes #973

The "booking request received" email sent to the booker now contains a link to the reservation page:

```
{EXCAL_URL}/booking/confirmed/eyJ...Gsw
```

where the JWT is the same `bookingConfirmationToken` that `POST /api/booking-links/{id}/book` returns, so the booker can review (and cancel) their reservation straight from the email instead of relying on the HTTP response of the booking call.

## Changes

- New `BookedEventLinkFactory` (calendar-api), mirroring `EventParticipationActionLinkFactory`: signs a `BookedEvent` and builds `{spa.excal.url}/booking/confirmed/{jwt}`. It uses the `participationActionLinks` (non-expiring) signer, since excal tokens are expected to outlive the default 12h `jwt.key.validity` — a booking may well be more than 12h away.
- `BookingCreated` now carries the `BookedEvent`, built once in `BookingLinkReservationService` and reused for both the notification and the HTTP response.
- `BookingLinkRequestAcknowledgementNotifier` exposes `content.reservationLink` to the pug model.
- Template `event-booking-request-received`: new link, translated in en/fr/ru/vi.

## Tests

Added `acknowledgementEmailShouldContainAWorkingLinkToTheReservationPage` in `BookingLinkReservationRouteTest`: asserts the email carries exactly one such link and that the JWT it embeds is accepted by `GET /api/booked-event`. It passes along with the pre-existing acknowledgement test.

---
*Generated automatically*


chibenwa (commented):
+1

vttranlina (approved):

PR #977, opened 2026-07-21
ISSUE-976 Event update: base not-email-in-the-past guard on the new event time

chibenwa:
Closes #976

## Problem

The not-email-in-the-past guard (esn-sabre `IMipPlugin::testIfEventIsExpired`) suppresses an update notification when the event has already started. In the scenario from the issue — Bob invites Alice at 9:00, Alice does not show up, then at 9:05 Bob reschedules the event to 10:00 — the attendee should still be notified because the *new* start time is in the future.

## Findings

In the async scheduling path (`ItipLocalDeliveryConsumer` → `ItipEmailNotificationPublisher` → `EventEmailConsumer`), which is the side-service port of esn-sabre `IMipPlugin::schedule`, the expiration guard was **not** wired: `CalendarEventUtils.vEventExpired` (ported from `testIfEventIsExpired` in #668) existed with tests but was never called. So notifications were published regardless of the event being past.

## Fix

Wire `CalendarEventUtils.vEventExpired` into `ItipEmailNotificationPublisher.buildNotificationMessages`, filtering out notifications whose event has expired. The check is based on the event each notification carries, which always reflects the **new (rescheduled)** time:

- reschedule of a past event to a future slot → still notifies attendees;
- an update whose new start is already in the past → suppressed.

This mirrors esn-sabre, where the guard is applied to every produced `eventMessage` in the publish loop, so it currently covers all methods (REQUEST/REPLY/CANCEL/COUNTER). Happy to restrict it to REQUEST-only if you prefer a narrower scope.

## Tests

- `rescheduleFromPastToFutureShouldStillNotify` — the issue scenario.
- `updateShouldBeSuppressedWhenNewEventIsAlreadyInThePast`.
- Existing `ItipEmailNotificationPublisherTest` / `ItipLocalDeliveryConsumerTest` fixtures use fixed 2026-04 dates, so they now run against a fixed clock anchored before those events; async integration tests use far-future (year 3025) events and keep the system clock.

---
*Generated automatically*

### DIFF
--- a/calendar-amqp/src/main/java/com/linagora/calendar/amqp/ItipEmailNotificationPublisher.java
+++ b/calendar-amqp/src/main/java/com/linagora/calendar/amqp/ItipEmailNotificationPublisher.java
@@ -21,6 +21,7 @@ package com.linagora.calendar.amqp;
 import static com.linagora.calendar.storage.event.EventParseUtils.createInstanceVEvent;
 
 import java.net.URI;
+import java.time.Clock;
 import java.time.LocalDate;
 import java.time.chrono.ChronoZonedDateTime;
 import java.time.temporal.Temporal;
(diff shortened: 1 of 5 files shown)


chibenwa (commented):
+1

PR #979, opened 2026-07-21
ISSUE-978 GET /api/team-calendars/{id}

chibenwa:
Closes #978

Adds `GET /api/team-calendars/{id}`, symmetrical to `GET /api/resources/{id}`.

## What it does

The endpoint returns the details of a single team calendar, mapping the data stored in database:

- `_id`, `name`, `displayName`
- `timestamps` (`creation` / `updatedAt`)
- the enriched `domain` representation (same shape as the resource endpoint, including domain administrators)

It reuses the existing `CrossDomainAccessControl` so that:
- any authenticated user of the domain can read a team calendar of that domain,
- the platform admin can read team calendars of any domain,
- a `404` is returned when the team calendar does not exist or belongs to a domain the caller may not access.

## Notes

The path is pluralized (`team-calendars`) to stay consistent with the existing `/api/resources/{id}` convention (the `team-calendar` object type from people search serializes in the singular, but REST collection paths are plural here).

## Tests

`TeamCalendarRouteTest` covers the nominal response body, cross-domain access control (user / admin / foreign domain), unknown id (`404`) and missing authentication (`401`).

---
*Generated automatically*

PR #981, opened 2026-07-21
ISSUE-980 GET /api/entity/{id}

chibenwa:
Closes #980

Adds a single `GET /api/entity/{id}` endpoint that aggregates the existing `GET /api/resources/{id}`, `GET /api/users/{id}`, `GET /api/domains/{id}` and `GET /api/team-calendars/{id}` lookups.

It resolves the type of an entity from its id and returns the matching entity wrapped under a discriminating key, reusing the DTOs already served by the dedicated endpoints:

- `{"user": {...}}`
- `{"resource": {...}}`
- `{"domain": {...}}`
- `{"teamCalendar": {...}}`

Cross-domain access control is preserved: a missing id, or an entity from a domain the authenticated user may not access, returns `404` (admins bypass the check, as elsewhere).

### Notes

- The team-calendar branch mirrors the response shape proposed in #979 (ISSUE-978). That PR is still open, so this route carries its own `TeamCalendarResponseDTO` for now; the two can be consolidated once #979 lands.
- `ResourceRoute` was refactored to expose a static `ResourceResponseDTO.from(...)` factory so the entity route reuses the exact same serialization.

### Acceptance

- [x] Implementation (`EntityRoute`, registered in `RestApiModule`)
- [x] Test class (`EntityRouteTest`, new class, 8 tests)
- [x] Documentation in `openpaasApi.md`

---
*Generated automatically*

### DIFF
--- a/calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/EntityRoute.java
+++ b/calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/EntityRoute.java
@@ -19,10 +19,14 @@
 package com.linagora.calendar.restapi.routes;
 
 import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.List;
 
 import jakarta.inject.Inject;
 import jakarta.inject.Named;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.james.jmap.Endpoint;
 import org.apache.james.jmap.http.Authenticator;
 import org.apache.james.mailbox.MailboxSession;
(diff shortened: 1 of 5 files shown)


calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/EntityRoute.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Override
+    Mono<Void> handleRequest(HttpServerRequest req, HttpServerResponse res, MailboxSession session) {
+        String id = req.param(ENTITY_ID_PATH_PARAM);
+        return asResourceEntity(id, session)
+            .switchIfEmpty(asUserEntity(id, session))
+            .switchIfEmpty(asTeamCalendarEntity(id, session))
+            .switchIfEmpty(asDomainEntity(id, session))
+            .switchIfEmpty(Mono.error(NotFoundException::new))
```
chibenwa: user
then team calendar
then resource
then domain

vttranlina (approved):

PR #992, opened 2026-07-28
ISSUE-990 Use DAV/Sabre ACL as source of truth for resolving resource administrators

vttranlina:
resolve https://github.com/linagora/twake-calendar-side-service/issues/990


chibenwa (approved):
Looks WAAAY simpler indeed!

PR #1000, opened 2026-07-31
ISSUE-999 Add name of extra attendees and resources to response of get request

hungphan227:


chibenwa (approved):

chibenwa (commented):
This change should be scoped to the public api only.

No change for booking link crud.

If that is already the case then this is cool.

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/BookingLinkListRoute.java
```diff
@@ -57,7 +59,7 @@ Endpoint endpoint() {
     @Override
     Mono<Void> handleRequest(HttpServerRequest request, HttpServerResponse response, MailboxSession session) {
         return bookingLinkDAO.findByUsername(session.getUser())
-            .map(BookingLinkDTO::from)
+            .concatMap(bookingLinkDTOFactory::create)
```
vttranlina, line 62: Performance issue: N+1 lookups when listing booking links
The current implementation enriches each booking link independently:

```
findByUsername(username)
          |
          v
   BookingLink 1
          |
          +-- resolve attendee 1
          +-- resolve attendee 2
          +-- resolve resource 1
          |
          v
   BookingLink 2
          |
          +-- resolve attendee 1     <-- duplicated lookup
          +-- resolve resource 1     <-- duplicated lookup
          +-- resolve resource 2
          |
          v
   BookingLink N
```

**Suggested approach**
First collect the booking links returned by findByUsername, extract all unique attendee and resource IDs, resolve them once, and then use the resulting maps to build every DTO

WDYT?

hungphan227: hmm it depends on the probability of attendee duplication. I don't think it is much

vttranlina (approved):

PR #1002, opened 2026-08-04
feat(meet): Create the room instead of hoping a random slug names one

mmaudet:
Part of linagora/twake-calendar-side-service#1004.

## The problem

The calendar frontend invents the video conference room code in the browser —
`generateMeetingId`, three then four then three random letters — and writes it
straight into `X-OPENPAAS-VIDEOCONFERENCE`. Nothing ever tells Meet about it.

On any deployment where Meet does not accept unregistered rooms, **every
generated link is dead**, for every client, web included:

```
$ curl -s -o - -w '%{http_code}' https://<meet-host>/api/v1.0/rooms/mjj-beyv-zai/
{"detail":"No Room matches the given query."}404
```

It also breaks the host delegation this branch builds on. `MeetHostDelegationService`
cannot grant host rights on a room that does not exist, and says so:

```
WARN  c.l.c.a.m.MeetHostDelegationService -
  Meet room not found for slug 'mjj-beyv-zai' (organizer <organiser@…>)
  — skipping delegates [<delegate@…>]
```

## Why the fix belongs here

Meet's external API already does exactly what is needed. `RoomViewSet` includes
`mixins.CreateModelMixin`, and `RoomSerializer.create` is written for this
caller:

```python
def create(self, validated_data):
    """Create room with secure defaults for application delegation."""
    validated_data["name"] = utils.generate_room_slug()
    ...
```

with `to_representation` adding `url` from `APPLICATION_BASE_URL`, and
`perform_create` granting the scoped user `OWNER`.

So `POST /external-api/v1.0/rooms/` mints the code itself and returns a finished
link, owned by the organiser. This service already holds the credentials and
already knows how to obtain an application token scoped to a user
(`MeetApplicationClient.fetchApplicationToken`), which is why the endpoint
belongs here rather than in the browser.

Note the constraint this implies: `RoomSerializer.Meta.read_only_fields` contains
`name` and `slug`, and `Room.clean_fields` does `self.slug = slugify(self.name)`.
**A caller cannot impose a slug.** Any fix that keeps the frontend's invented
code would require changing Meet; asking Meet for the code does not.

## What this adds

- `MeetApplicationClient.createRoom(token)` — `POST` on the existing `ROOMS_PATH`,
  reads `url` from the response. Fails loudly and names `APPLICATION_BASE_URL`
  when that field is absent, rather than returning a half-built link that would
  end up as an empty href in an invitation.
- `POST /api/videoconference` (`VideoConferenceRoute`) — authenticates as every
  other route does, resolves the caller's address, and returns `{"url": "…"}`.
  Answers `404` when Meet is not configured, so a caller can tell "this
  deployment does not mint rooms" from "minting failed" and fall back to its own
  link generation; `502` when Meet refuses, so the failing side is named.
- `meet.room.access_level`, optional. Meet's `EXTERNAL_API_DEFAULT_ACCESS_LEVEL`
  is `trusted`, which only admits authenticated users of the instance — **an
  external guest invited to a meeting cannot get in, and the organiser being
  present changes nothing.** Deployments that invite guests want `public` here,
  which Meet gates behind its own `EXTERNAL_API_ALLOW_PUBLIC_ACCESS`. Leaving the
  property unset keeps Meet's default, so nothing changes for deployments that
  do not set it.

## Measured on a Twake deployment

| step | result |
| --- | --- |
| two video conferences created from the calendar | two rooms appear in Meet, 15 s apart |
| `GET /api/v1.0/rooms/<slug>/` | `200` for both — `404` before |
| host delegation | `Granted admin on Meet room … (HTTP 201)`, where it logged `Meet room not found` the day before |
| mobile client | reaches its prejoin screen instead of "not found on the server" |

The delegation check is the strongest of the four and it is indirect:
`MeetHostDelegationService` resolves the room **by the slug it reads from the
ICS**. That it finds one proves the event carries the created room's slug, not
merely that a room was created around the right time.

Host delegation was never broken — it was blocked by a room that did not exist.
This repairs it as a consequence rather than working around it.

## Tests

Six on the client, run against WireMock, including **both polarities** of the
access-level branch: a configured level is sent, an unconfigured one leaves the
choice to Meet. Verified by mutation — making `createRoom` always send a level
turns exactly that second test red.

The route is thin glue over the tested client and is not covered here; it needs
the app-level integration harness.

## Companion change

The frontend half is
linagora/twake-calendar-frontend#1247. Either alone is a no-op: the frontend
falls back to local generation on `404`, and this route is never called until the
frontend asks.

## Base branch

Opened against `feat/meet-host-delegation`, which this builds on and which is not
yet merged.

### DIFF
--- a/calendar-amqp/src/main/java/com/linagora/calendar/amqp/meet/MeetApplicationClient.java
+++ b/calendar-amqp/src/main/java/com/linagora/calendar/amqp/meet/MeetApplicationClient.java
@@ -129,6 +129,44 @@ public class MeetApplicationClient {
     }
 
     /**
+     * Ask Meet to create a room owned by the token's scoped user, and return
+     * its public URL.
+     *
+     * <p><strong>The caller does not choose the room code.</strong> Meet's
+     * external API marks {@code name} and {@code slug} read-only and its
+     * serializer does {@code validated_data["name"] = utils.generate_room_slug()},
+     * so the code is Meet's to mint. That is the whole point: a code invented
+     * anywhere else names a room that does not exist.
+     *
+     * <p>{@code perform_create} grants the scoped user {@code OWNER}, so the
+     * organiser gets host controls — lobby admission, recording — which a
+     * pre-minted link never gave them.
+     */
+    public Mono<String> createRoom(String bearerToken) {
+        ObjectNode body = MAPPER.createObjectNode();
+        config.roomAccessLevel().ifPresent(level -> body.put("access_level", level));
+        byte[] payload = serialize(body);
+
+        return client.headers(h -> {
+                h.set(HttpHeaderNames.AUTHORIZATION, "Bearer " + bearerToken);
+                h.set(HttpHeaderNames.CONTENT_TYPE, HttpHeaderValues.APPLICATION_JSON);
+            })
+            .post()
+            .uri(ROOMS_PATH)
+            .send(Mono.just(Unpooled.wrappedBuffer(payload)))
+            .responseSingle((response, bodyMono) -> {
+                HttpResponseStatus status = response.status();
+                return bodyMono.asString().defaultIfEmpty("").flatMap(bodyString -> {
+                    if (status.code() >= 200 && status.code() < 300) {
+                        return extractRoomUrl(bodyString);
+                    }
+                    return Mono.error(new MeetApiException(
+                        "Failed to create room: HTTP " + status.code() + " — " + bodyString));
+                });
+            });
+    }
+
+    /**
      * Grant admin (or member) access on {@code roomId} to {@code delegateEmail}.
      * Idempotent by design of the server (see Meet Patch A).
      */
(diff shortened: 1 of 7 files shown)


calendar-amqp/src/main/java/com/linagora/calendar/amqp/meet/MeetApplicationClient.java
```diff
@@ -58,8 +58,10 @@ public class MeetApplicationClient {
     private static final String GRANT_ACCESS_URL_TEMPLATE = "/external-api/v1.0/rooms/%s/grant-access/";
 
     private final HttpClient client;
+    private final MeetConfiguration config;
 
     public MeetApplicationClient(MeetConfiguration config) throws SSLException {
+        this.config = config;
         this.client = config.enabled() ? httpClientFor(config) : null;
     }
```
chibenwa: A module chooser on MeetConfiguration::enabled would ensure meet related stuff is injected if and only if meet is enabled, we'd save on runtime guards that are hard to debug

calendar-rest-api/src/main/java/com/linagora/calendar/restapi/routes/VideoConferenceRoute.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * <p>Asking Meet for the room fixes that at the source and buys two things a
+ * pre-minted code could never give: the room exists before the invitation is
+ * sent, and the organiser is its {@code OWNER}, so lobby admission, recording
+ * and the rest of the host controls actually work.
+ *
+ * <p>The Meet application credentials stay server-side, which is why this is a
+ * route here rather than a call from the browser.
+ */
+public class VideoConferenceRoute extends CalendarRoute {
```
chibenwa: I don't understand why in this PR we ask for the client to explicitly create the room through a HTTP call.

CF https://github.com/linagora/twake-calendar-side-service/pull/1001 is working through asynchronous event listener for delegates. I'd prefer we create the room as a preamble of that listener with the prior condition to have a `X-OPENPAAS-VIDEOCONFERENCE`.

Benefits: 
 - No new API surface, no client code changes
 - Transparent for **Booking Links** where we do not have a client that calls this route for us. This would naturally factorize the code, which imo is elegant.

Cc @Crash--

calendar-rest-api/pom.xml
```diff
@@ -33,15 +33,22 @@
...
+            That client speaks HTTP, not AMQP, and sits in this module because the
+            host-delegation consumer was its first caller; a follow-up could pull
+            the two classes into a module of their own rather than have the REST
+            layer depend on the messaging one.
+        -->
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>calendar-storage-api</artifactId>
+            <artifactId>calendar-amqp</artifactId>
         </dependency>
```
chibenwa: Careful with dependency ordering 
We can get rid of the verbose comment.

vttranlina:
alternative https://github.com/linagora/twake-calendar-side-service/pull/1086

Issue #989, opened 2026-07-24
Common contact: OUTBOUND and OUTBOUND


chibenwa:
CF https://github.com/linagora/twake-workplace-private/pull/1176

The side service needs to expose a convenience service in order to over apps to integrate easily.

Reuse the rabbitmq connection and the vhost of the platform (namely settings, subscription, etc...)

Turn this on as soon as the platform (settings) is configured.

## INBOUND

Exchange: `twake:contacts:collected`

Routing key: `` (intentionally empty)

```
{
  "userEmail": "btellier@linagora.com",
  "collectedContacts": [[ ... https://www.rfc-editor.org/info/rfc9553/ caljsondata ], ... other contacts ... ]
}
```

If ommitted we recompute the UID field from the `mailaddress` (same algorithm than twake mail - see below) and fall back to matrix id, then to phone number. No two contact with same uid.

## OUTBOUND

We advertize addition, uppdate and deletion to over apps.

Exchange: `twake:contacts:common`

Routing key: `` (intentionally empty)

```
{
   "audience":{"user":"btellier@linagora.com"}, // "audience":{"domain":"linagora.com"},
    "action": "ADD" // "DELETE", "UPDATE",
    "path": "addressbooks/PRINCIPAL/COLLECTION",
     "uid": "xxxxxx",
     "payload":  : [[ ... https://www.rfc-editor.org/info/rfc9553/ caljsondata ]]
}
````

### UID

The UID must be in order not to create duplicates the SHA-1 hash as a string of the main email address. Java example:

```
  public static DavUid createContactUid(MailAddress email) {
      return new DavUid(Hashing.sha1()
          .hashString(email.asString(), StandardCharsets.UTF_8)
          .toString());
  }
```

vttranlina:
Should the Common Contacts publisher reuse [TWPCommonRabbitMQConfiguration](https://github.com/linagora/twake-calendar-side-service/blob/master/saas/saas-rabbitmq/src/main/java/com/linagora/calendar/saas/TWPCalendarSubscriptionModule.java#L51) and its existing `twp.rabbitmq.uri` / `twp.rabbitmq.management.uri` settings?

Do we need separate enable/disable toggles for inbound and outbound Common Contacts, or should enabling `TWPCalendarSubscriptionModule` automatically enable both?

chibenwa:
[...]

PR #1017, opened 2026-08-19
ISSUE-989 Common contact: INBOUND

vttranlina:
ref https://github.com/linagora/twake-calendar-side-service/issues/989


saas/saas-rabbitmq/src/main/java/com/linagora/calendar/saas/contact/CollectedContactConverter.java
```diff
@@ -0,0 +1,150 @@
...
+            throw e;
+        } catch (CardException | RuntimeException | IOException e) {
+            throw new CollectedContactConversionException("""
+                Unable to convert collected JSContact to vCard:
+                %s
+                """.formatted(contactData.toPrettyString()), e);
+        }
+    }
+
+    static final class ContactUidResolver {
```
chibenwa: Is it a resolver?
I'd be tempted to call it `UidCalculator` or uid normalizer

uidNormalizer.generateNormalizedUid(card) 

?

saas/saas-rabbitmq/src/test/java/com/linagora/calendar/saas/contact/CollectedContactUidResolverTest.java
```diff
@@ -0,0 +1,323 @@
...
+            .hasMessage("Cannot generate contact UID: missing email, Matrix ID and phone number");
+    }
+
+    @SuppressWarnings("deprecation")
+    private static String sha1(String value) {
+        return Hashing.sha1()
+            .hashString(value, StandardCharsets.UTF_8)
+            .toString();
+    }
+}
```
chibenwa: I'd like to add one more test in here to ensure we allocate the same uid for email and for matrixId

btellier@linagora.com would get treated as @btellier:linagora.com 

and the two contacts merged.

saas/saas-rabbitmq/src/test/java/com/linagora/calendar/saas/contact/CollectedContactsConsumerTest.java
```diff
@@ -0,0 +1,478 @@
...
+                    EMAIL;PROP-ID=personal:personal@example.com
+                    EMAIL;PROP-ID=work:work@example.com
+                    SOCIALPROFILE;SERVICE-TYPE=matrix;PROP-ID=matrix;VALUE=text:@alice:matrix.e
+                     xample.com
+                    END:VCARD
+                    """)));
+    }
+
+    @Test
+    void shouldUpdateExistingContactWithSameUid() throws Exception {
```
chibenwa: Add a test ensuring that...
 - message 1: email address only
 - message 2: matrix id only

Expected outcome: collected contact have both fields (merged in place)

vttranlina: I just checked: esn-sabre does not provide a parameter or API supporting vCard merging. 
The libraries currently used in the side service do not provide a merge API either.

It looks like, before each upsert, we will need to retrieve the existing contact. If it exists, we apply the merge logic and then PUT (upsert) the merged vCard back to Sabre.

chibenwa: Yes !

saas/saas-rabbitmq/src/main/java/com/linagora/calendar/saas/contact/CollectedContactConverter.java
```diff
@@ -67,6 +75,75 @@ public ConvertedContact convert(ObjectNode contactData) {
...
+    }
+
+    private boolean shouldMergeComplementaryIdentities(Card existingCard, Card incomingCard) {
+        boolean existingHasEmail = uidCalculator.emailAddress(existingCard).isPresent();
+        boolean existingHasMatrixId = uidCalculator.matrixId(existingCard).isPresent();
+        boolean incomingHasEmail = uidCalculator.emailAddress(incomingCard).isPresent();
+        boolean incomingHasMatrixId = uidCalculator.matrixId(incomingCard).isPresent();
+
+        return existingHasEmail && incomingHasMatrixId && !incomingHasEmail
+            || existingHasMatrixId && !existingHasEmail && incomingHasEmail && !incomingHasMatrixId;
```
chibenwa: This condition is **way** to complex.

Break it down into smaller boolean expressions in order to make it readable.

hungphan227 (approved):

chibenwa (approved):
Acceptable to me.

We will handle edge cases in follow up tickets

PR #1019, opened 2026-08-20
ISSUE-1005 ItipEmailNotificationPublisher should send email notification to booker when organizer decline event

hungphan227:
This update follows https://github.com/linagora/esn-sabre/pull/458 to fully allow sending email notification to booker when organizer decline event


chibenwa (approved):
Fine with me

PR #1022, opened 2026-08-21
Set distinct User-Agent headers for DAV clients

vttranlina:
## Why

Make it easy to distinguish direct calendar-side-service DAV traffic  when debug log in esn-sabre

## Expected esn-sabre logs

```text
DEBUG User-Agent: twake-calendar-side-service ReactorNetty/<version>
DEBUG User-Agent: twake-calendar-dav-proxy ReactorNetty/<version>
```


chibenwa (approved):
Very nice idea.

I'll promote a PR to do this too with Twake Mail

